PR #212, opened 2016-04-29
[BT-653]: Change deploy task to deploy to GitHub Pages instead of AWS. 

cjcenizal:
- Migrate to react-router-redux.
- Update CSS and JS paths to be relative to index file.
- Add history dependency back, and switched to hash history, since we won't have control over the back-end with GitHub Pages.

You can check out the deployment result at http://smaato.github.io/ui-framework/


gulpfile.js
```diff
@@ -30,12 +31,18 @@ gulp.task('copySource', gulpTasks.copy({
...
 }).task);
 
-gulp.task('deploy', gulpTasks.deploy({
-  accessKeyId: process.env.AWS_ACCESS_KEY_ID || 'No ENV',
-  bucketName: process.env.AWS_BUCKET_UI_FRAMEWORK || 'No ENV',
-  secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY || 'No ENV',
-  src: `${DISTRIBUTION_DIR}/**/*.*`,
-}).task);
+gulp.task('deployToGitHubPages', () => (
+  gulp.src('./dist/**/*')
```
jenslonkowski: We should use the `DISTRIBUTION_DIR` constant here.

gulpfile.js
```diff
@@ -30,12 +31,18 @@ gulp.task('copySource', gulpTasks.copy({
   src: `${SOURCE_DIR}/guide/**/*.jsx`,
 }).task);
 
-gulp.task('deploy', gulpTasks.deploy({
-  accessKeyId: process.env.AWS_ACCESS_KEY_ID || 'No ENV',
-  bucketName: process.env.AWS_BUCKET_UI_FRAMEWORK || 'No ENV',
-  secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY || 'No ENV',
-  src: `${DISTRIBUTION_DIR}/**/*.*`,
-}).task);
+gulp.task('deployToGitHubPages', () => (
```
jenslonkowski: What do you think of calling this task `deploy` and...

gulpfile.js
```diff
@@ -30,12 +31,18 @@ gulp.task('copySource', gulpTasks.copy({
...
+  gulp.src('./dist/**/*')
+    .pipe(ghPages())
+));
+
+gulp.task('deploy', done => {
```
jenslonkowski: ...calling this task `buildAndDeploy`?

jenslonkowski:
What do you think of adding the link to the README?

jenslonkowski:
Great work! Just some small comments and then 👍

gulpfile.js
```diff
@@ -30,12 +31,18 @@ gulp.task('copySource', gulpTasks.copy({
   src: `${SOURCE_DIR}/guide/**/*.jsx`,
 }).task);
 
-gulp.task('deploy', gulpTasks.deploy({
-  accessKeyId: process.env.AWS_ACCESS_KEY_ID || 'No ENV',
-  bucketName: process.env.AWS_BUCKET_UI_FRAMEWORK || 'No ENV',
-  secretAccessKey: process.env.AWS_SECRET_ACCESS_KEY || 'No ENV',
-  src: `${DISTRIBUTION_DIR}/**/*.*`,
-}).task);
```
jenslonkowski: I suggest to keep this task as `deployToAws`.

gulpfile.js
```diff
@@ -30,12 +31,18 @@ gulp.task('copySource', gulpTasks.copy({
...
+    .pipe(ghPages())
+));
+
+gulp.task('deploy', done => {
+  runSequence(
+    'production',
+    'deployToGitHubPages',
+    done
+  );
+});
```
jenslonkowski: I think this task reduces the flexibility for deployments (especially if we decide to keep the AWS task as well). What's the benefit?

ahollesen:
Looking good 👍

gulpfile.js
cjcenizal: What do you mean by reducing flexibility for deployments? This is just a convenience task for deploying the production build, since you can just run the 1 task instead of having to run production, wait for it to complete, and then run a separate task. And you can't mess it up by forgetting to run production first. :)

jenslonkowski: But sometimes you don't want to run production first, e.g. when you want to deploy the unminified scripts or when you want to run production first, then do end-to-end tests and then deploy without running production again or when you want to deploy to AWS instead. You can always run `gulp production && gulp deployToGitHubPages` instead of having this task. We can keep it but I don't think it's that useful.

PR #213, opened 2016-05-03
[BT-652]: Improve Home page and Navigation.

cjcenizal:
image

image

### DIFF
--- a/src/guide/components/navigation/Navigation.jsx
+++ b/src/guide/components/navigation/Navigation.jsx
@@ -17,50 +17,42 @@ export default class Navigation extends Component {
     super(props);
   }
 
+  renderNavMenu(name, routeItems) {
+    // Add items.
+    const navItems = routeItems.map((route, index) => (
+      <NavItem
+        href={route.href}
+        path={route.path}
+        onClick={this.props.onClickNavItem}
+        key={`${name}${index}`}
+      >
+        {route.name}
+      </NavItem>
+    ));
+
+    // Add title at the top.
+    navItems.unshift(
+      <NavTitle key={name}>
+        {name}
+      </NavTitle>
+    );
+
+    return (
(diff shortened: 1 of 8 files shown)


src/guide/components/navigation/Navigation.jsx
```diff
@@ -15,80 +15,73 @@ export default class Navigation extends Component {
...
-          </NavButton>
+          </NavItem>
         );
       }
-    }.bind(this);
+    };
 
-    const navItems = [];
+    const componentNavItems = [];
+    addRoutesToNav('Components', this.props.componentRoutes, componentNavItems);
```
jenslonkowski: Please change this to multi-line to match lines 43-48.

jenslonkowski:
Great improvements! 👍

ahollesen:
Yup, looks nice 👍

src/guide/components/navigation/Navigation.jsx
cjcenizal: Sure, but lines 43-48 were only changed to adhere to the linter.

jenslonkowski: I see. But I think in general it's a good idea to use multiple lines if there are multiple arguments. Would you agree?

cjcenizal: It depends on the situation for me. I think formatting this to match 43-48 is a good idea.

cjcenizal:
Thanks guys!

plain:
CJ - the design looks great. I may want to tweak it in the future, but I am fine with this to get it out the door.

cjcenizal:
@plain Awesome! Thanks for taking a look. I'm sure you can really pimp it out... I didn't spend a lot of time on the look and feel.

PR #214, opened 2016-05-03
[BT-656]: Fix ModalStack so that it doesn't crop overflowing content.

cjcenizal:
Update example to demo this.

### DIFF
--- a/src/guide/views/dropdown/DropdownExample.jsx
+++ b/src/guide/views/dropdown/DropdownExample.jsx
@@ -26,7 +26,7 @@ export default class DropdownExample extends Component {
     }, {
       name: 'Berry',
     }, {
-      name: 'Chippewa',
+      name: 'Corn',
     }, {
       name: 'Daffodil',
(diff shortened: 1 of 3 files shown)


src/guide/views/dropdown/DropdownExample.jsx
```diff
@@ -26,7 +26,7 @@ export default class DropdownExample extends Component {
     }, {
       name: 'Berry',
     }, {
-      name: 'Chippewa',
+      name: 'Corn',
```
cjcenizal: Chippewa is the name of a Native American tribe, but everyone else is something edible... one of these things was not like the others. 😮

ahollesen: Technically daffodils are edible, but more a flower than food right?! ;)

ahollesen:
Looking good 👍

jenslonkowski:
Great! 👍

PR #215, opened 2016-05-03
[BT-655]: Convert all children props to PropType.any.

cjcenizal:
There's no advantage to restricting children type.

### DIFF
--- a/src/framework/label/Label.jsx
+++ b/src/framework/label/Label.jsx
@@ -25,7 +25,7 @@ const Label = props => {
 };
 
 Label.propTypes = {
-  children: PropTypes.string.isRequired,
+  children: PropTypes.any,
   className: PropTypes.oneOfType([
     PropTypes.object,
     PropTypes.array,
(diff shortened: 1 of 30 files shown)


ahollesen:
Yup 👍

src/framework/label/Label.jsx
```diff
@@ -25,7 +25,7 @@ const Label = props => {
 };
 
 Label.propTypes = {
-  children: PropTypes.string.isRequired,
+  children: PropTypes.any,
```
jenslonkowski: Shouldn't the children still be required?

jenslonkowski:
Good job CJ! I have one question and then: 👍

src/framework/label/Label.jsx
cjcenizal: I think the same line of thought applies... what advantage does requiring a child have? I _think_ we should only use isRequired for props that will cause the component's internal logic to break if they're missing. Though the official documentation isn't clear on this and I can't find much written on the subject.

An advantage of using isRequired more sparingly is that it will make our unit tests less verbose, since we won't need to provide as many no-op props for each test case.

PR #217, opened 2016-05-04
[BT-657]: Truncate text when Dropdown is narrower than its label.

cjcenizal:
- Refactor ColumnLayout to use inline-block instead of flexbox.

image

image


jenslonkowski:
Much great! Such wow! 👍

ahollesen:
Looking good 👍

PR #218, opened 2016-05-10
[BT-658] Adjust components for InactiveUserView

ahollesen:
- reusing background-color for body
- added isTargetBlank prop to link + example

### DIFF
--- a/src/framework/link/Link.jsx
+++ b/src/framework/link/Link.jsx
@@ -10,18 +10,13 @@ const Link = props => {
     }
   }
 
-  let target;
-  if (props.isTargetBlank) {
-    target = '_blank';
-  }
-
   return (
     <a
       className="link"
       href={props.href}
       onClick={onClick}
       data-id={props.dataId}
-      target={target}
+      target={props.target}
     >
       {props.children}
     </a>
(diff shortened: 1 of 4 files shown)


jenslonkowski:
Spotless! 👍

src/framework/link/Link.jsx
```diff
@@ -28,6 +34,7 @@ Link.propTypes = {
   href: PropTypes.string,
   onClick: PropTypes.func,
   data: PropTypes.any,
+  isTargetBlank: PropTypes.bool,
```
cjcenizal: I think we should leverage convention for components which are simple wrappers around HTML elements. So I think we should just make this `target` instead, and require the dev to pass in `"_blank"`.

cjcenizal:
1 comment, then 👍

src/framework/link/Link.jsx
ahollesen: Actually that was my first version, but then I thought that any other than `"_blank"` wouldn't be a proper option, as they are rarely or actually never used. I'm fine with both, the call is yours :)

cjcenizal: I hear you, I think we should go with my suggestion. At some point we might want to target specific tabs (http://stackoverflow.com/questions/6903815/how-to-open-link-url-in-specific-browser-tab-window-page).

PR #223, opened 2016-05-12
[BT-686]: Add CSS dependency upon published icon-library CSS.

cjcenizal:
This will allow us to develop the UI Framework (and will let others view it) with icons from the icon-library, without violating the prohibition against distribution. My reasoning is that since we can share the icons to our BT users via a CSS file, we can share the icons elsewhere via a CSS file too.

### DIFF
--- a/src/guide/index.jade
+++ b/src/guide/index.jade
@@ -21,6 +21,13 @@ html(lang='en')
       type = "text/css"
     )
 
+    //- Icons
+    link(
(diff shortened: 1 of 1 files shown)


jenslonkowski:
I think we need to verify your reasoning. Have you read the licence or talked with legal?

ahollesen:
I think we should get legal confirmation here

cjcenizal:
Confirmed with @plain.

PR #224, opened 2016-05-12
[BT-685]: Rename Sorter.simpleSort to sortBy.

cjcenizal:
- Improve sort method to accept a property name as a value provider.

### DIFF
--- a/src/framework/services/sort/Sorter.js
+++ b/src/framework/services/sort/Sorter.js
@@ -10,37 +10,50 @@ function normalizeValue(value) {
 
 function sort(
   originalItems,
-  valueProviderOrProviders,
+  propertyNameOrValueProviderOrProviders,
   providerPropertyOrIndex,
   isDescending = false
 ) {
-  const isSingleValueProvider = typeof valueProviderOrProviders === 'function';
+  const isPropertyName =
+    typeof propertyNameOrValueProviderOrProviders === 'string';
 
-  // If we've received an array of value providers, then we need to make sure
-  // an acceptable providerPropertyOrIndex has been provided.
-  if (!isSingleValueProvider) {
-    const typeOfProviderPropertyOrIndex = typeof providerPropertyOrIndex;
-    if (typeOfProviderPropertyOrIndex !== 'string' &&
-      typeOfProviderPropertyOrIndex !== 'number') {
-      throw new Error(
-        `providerPropertyOrIndex must be a string or number.
-        Got ${providerPropertyOrIndex}.`
-      );
+  const isSingleValueProvider =
+    typeof propertyNameOrValueProviderOrProviders === 'function';
+
+  if (!isPropertyName) {
+    // If we've received an array of value providers, then we need to make sure
+    // an acceptable providerPropertyOrIndex has been provided.
+    if (!isSingleValueProvider) {
+      const typeOfProviderPropertyOrIndex = typeof providerPropertyOrIndex;
+      if (typeOfProviderPropertyOrIndex !== 'string' &&
+        typeOfProviderPropertyOrIndex !== 'number') {
+        throw new Error(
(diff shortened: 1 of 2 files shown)


jenslonkowski:
I think putting all the logic inside one method is too hard to follow. I would prefer having smaller methods that each hold small portions of the logic:
- `sort` only accepts an array of valueProviders
- `simpleSort` only accepts a single valueProvider and puts it in an array to call `sort`
- `sortBy` only accepts a string and creates a valueProvider to call `simpleSort`

ahollesen:
I think Jens' suggestion would make it slightly easier to use, so I would prefer, but I'd be fine with the current state, so 👍

cjcenizal:
@jenslonkowski Great point about the code being difficult to follow. I think your suggestion means reducing functionality though (because you couldn't use `sort` to sort against a property name). I'm also concerned because the interface because more complex -- it's hard to tell the functions apart by their names.

Conceptually, I think we only need two functions: one "master" sorting function (which will mostly be used for sorting along multiple axes, e.g. sorting a table by different columns), and one convenience function for sorting along a single axis (e.g. sort an array of items by name).

I think the code is difficult to follow because of the interrelated (yet spatially separate) conditionals within the sort function. I've refactored the code to use a single switch statement instead of multiple conditional statements. I've also abstracted the concept of a "value provider", which should reduce the mental overhead when reading the sort function logic.

src/framework/services/sort/Sorter.spec.js
```diff
@@ -244,13 +267,13 @@ describe('Sorter', () => {
...
-              a: item => item.a,
-              b: item => item.b,
+          it('defines a provider via a key', () => {
+            const valueProvidersObject = {
+              provider1: item => item.a,
+              provider2: item => item.b,
             };
             const sortedItems =
-              Sorter.sort(items, valueProvidersByProperty, 'b');
+              Sorter.sort(items, valueProvidersObject, 'provider2');
```
cjcenizal: Just renamed variables in this test to make it clearer.

src/framework/services/sort/Sorter.spec.js
```diff
@@ -137,7 +160,7 @@ describe('Sorter', () => {
       describe('isDescending parameter', () => {
         describe('when false (default)', () => {
           it('sorts items ascending', () => {
-            const sortedItems = Sorter.sort(items, valueProvidersByIndex, 0);
+            const sortedItems = Sorter.sort(items, valueProvidersArray, 0);
```
cjcenizal: Renamed for clarity.

src/framework/services/sort/Sorter.spec.js
```diff
@@ -63,39 +63,39 @@ describe('Sorter', () => {
...
-        Sorter.simpleSort(items, valueProvider);
+        Sorter.sortBy(items, valueProvider);
         expect(Sorter.sort)
           .toHaveBeenCalledWith(items, valueProvider, undefined, undefined);
       });
     });
 
     describe('sort method', () => {
-      const valueProvidersByIndex = [
+      const valueProvidersArray = [
```
cjcenizal: Renamed for clarity.

src/framework/services/sort/Sorter.js
```diff
@@ -1,5 +1,60 @@
...
+      throw new Error(
+        `providerKeyOrIndex must be a string or number.
+        Got ${providerKeyOrIndex}.`
+      );
+    }
+  }
+
+  getValueFor(item) {
+    // NOTE: For now (for the sake of simplicity), we're assuming the providers
+    // are functions, not property names.
```
ahollesen: This could be checked inside the constructor as well.

ahollesen:
One small comment, else 👍

jenslonkowski:
Great improvements! 👍

PR #225, opened 2016-05-13
[BT-673]: Add PanelProgress component.

cjcenizal:


jenslonkowski:
Looks good! But according to the examples in https://github.com/SmaatoUI/loading-indicator-prototypes the loading indicator for the lists should be smaller. Maybe we can adjust the Progress component to support this through a property.

ahollesen:
Agree with Jens

src/framework/body/BodyPanelItem.spec.jsx
```diff
@@ -25,12 +22,9 @@ describe('BodyPanelItem', () => {
...
 
-          const testCase = TestCaseFactory.createFromFunction(
-            BodyPanelItem,
-            props
-          );
-          expect(testCase.dom.getAttribute('class')
-            .indexOf('bodyPanelItem--largeRhythm') !== -1).toBe(true);
+          const testCase = TestCaseFactory.create(BodyPanelItem, props);
+          expect(testCase.dom.className)
+            .toContain('bodyPanelItem--largeRhythm');
```
cjcenizal: I noticed these tests used an older convention, so I updated them.

src/guide/components/navigation/_navigation.scss
```diff
@@ -9,7 +9,7 @@
   background: #7b8396;
   width: 400px;
   overflow-y: auto;
-  transition: transform 0.5s ease-out;
+  transition: transform 0.25s ease-out;
```
cjcenizal: This was moving too slowly for me.

cjcenizal:
@jenslonkowski @ahollesen Good catch guys! Please re-review.

ahollesen:
Yup 👍

jenslonkowski:
Great! 👍

PR #227, opened 2016-05-13
[BT-694]: Limited width for grid body cells

jenslonkowski:
- Improved consistency of grid styles and a little cleanup
- Included max-width for body cells
- Adjusted flex of PickedSummary label to avoid a shrinking icon when label is truncated

### DIFF
--- a/src/framework/grid/footer/_gridFooterCell.scss
+++ b/src/framework/grid/footer/_gridFooterCell.scss
@@ -1,21 +1,10 @@
+
 .grid__footer__cell {
+  @include gridCellPadding;
   vertical-align: middle;
   line-height: 30px;
 }
 
   .grid__footer__cellLiner {
-    padding: 0 8px;
     height: 31px;
-  }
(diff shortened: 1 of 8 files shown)


cjcenizal:
Can you update the example to exhibit the problem we're solving in BT? Otherwise we can't test this code. 😄

src/framework/grid/footer/_gridFooterCell.scss
```diff
@@ -1,21 +1,19 @@
+
 .grid__footer__cell {
+  padding-left: 8px;
+  padding-right: 8px;
```
ahollesen:
```
padding-left: 8px;
padding-right: 8px;
```
is being used in `_gridHeader`, `_gridBodyCell`, `_gridFooterCell` and `_stickyGrid`. Could be pulled out.

ahollesen:
I think this should do. One small suggestion, else looking good 👍

PR #229, opened 2016-05-18
Consistent header padding for all screens

jenslonkowski:
* Padding of AppHeader and ViewHeader components isn't removed on screens wider then 1320px


ahollesen:
Looking good 👍

PR #232, opened 2016-05-26
[BT-745]: Deprecate GridRowRecycler in favor of generic RecycledList component.

cjcenizal:
- Add example of recycling list items with Menu.
- Add GridFakeRow component and ScrollPosition service.
- GridIcon passes event to onClick callback.
- Improve Grid and GridBody tests.

## Breaking changes

* Grid must be passed a GridBody as a child now; it no longer instantiates a GridBody itself.
* Grid no longer has columnsCount, rows, lazyLoadRows, initialLoadingRow, emptyRow, loadingRow, rowRecycler, or classBody props.
* GridBody no longer has columnsCount, firstRecycledRowOffset, or lastRecycledRowOffset props.


src/framework/grid/Grid.spec.js
```diff
@@ -103,21 +67,7 @@ describe('Grid', () => {
         };
 
         const testCase = TestCaseFactory.createFromElement(<Grid {...props} />);
```
jenslonkowski: We could switch to the `create` method everywhere in this file.

src/framework/grid/body/GridBody.spec.jsx
```diff
@@ -144,14 +69,14 @@ describe('GridBody', () => {
           columnsCount: 2,
           firstRecycledRowOffset: 1,
           lastRecycledRowOffset: 2,
-          loadingRow: <tr id="loadingRow"><td></td></tr>,
+          loadingRow: <tr className="loadingRow"><td></td></tr>,
         };
 
         const testCase = TestCaseFactory.createFromElement(
```
jenslonkowski: Same in this file.

src/framework/grid/body/GridFakeRow.spec.jsx
```diff
@@ -0,0 +1,52 @@
...
+    describe('style', () => {
+      it('is added as style to inner div', () => {
+        const props = {
+          columnsCount: 0,
+          style: {
+            backgroundColor: 'red',
+          },
+        };
+
+        const testCase = TestCaseFactory.createFromElement(
```
jenslonkowski: Same in this file.

src/framework/recycledList/RecycledList.jsx
```diff
@@ -0,0 +1,220 @@
...
+    this.updateItemPositions();
+  }
+
+  componentDidMount() {
+    this.props.scrollPosition.addListener(this.onScroll);
+  }
+
+  componentWillReceiveProps(nextProps) {
+    // When we receive more items, we need to update the recycled rows.
+    if (nextProps.items !== this.props.items) {
```
jenslonkowski: Shouldn't the length be checked here - is this direct comparison actually working?

src/framework/recycledList/RecycledList.jsx
```diff
@@ -0,0 +1,220 @@
...
+      this.updateItemPositions(nextProps.items);
+    }
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    if (nextState.firstItemOffset !== this.state.firstItemOffset) {
+      return true;
+    }
+
+    if (nextProps !== this.props) {
```
jenslonkowski: Same question here - can this evaluate to false?

src/framework/recycledList/RecycledList.jsx
```diff
@@ -0,0 +1,220 @@
...
+        firstItemOffset += itemDistanceScrolled;
+      }
+    } else {
+      // Calculate the position of the top edge of the first visible item.
+      // This is relevant if we are scrolling up and items are appearing at the
+      // top egde of the grid.
+      const firstItemTopEdgePosition = oldFirstItemOffset;
+
+      if (offsetScrollPosition < firstItemTopEdgePosition) {
+        // If we've scrolled the frist visible item just a tiny bit below the
```
jenslonkowski: *first

src/framework/services/scroll/ScrollPosition.js
```diff
@@ -0,0 +1,65 @@
...
+
+        // Call listeners.
+        this.listeners.forEach(listener => listener());
+      }
+    );
+  }
+
+  teardown() {
+    this.eventDispatcher.teardown();
+    this.listeners = undefined;
```
jenslonkowski: Why undefined and not an empty array?

src/framework/recycledList/RecycledList.jsx
```diff
@@ -0,0 +1,220 @@
...
+  recycledItemsCount: PropTypes.number.isRequired,
+  itemHeightProvider: PropTypes.func.isRequired,
+  // NOTE: Expects an instance of ScrollPosition.
+  scrollPosition: PropTypes.object.isRequired,
+};
+
+RecycledList.defaultProps = {
+  // NOTE: fakeItemElement needs to accept style as a prop.
+  fakeItemElement: <div />,
+};
```
jenslonkowski: I expected that the `RecycledList` would have a `componentWillUnmount` method that calls the `teardown` method of `scrollPosition` or did you decide against this tight coupling?

src/framework/services/scroll/ScrollPosition.js
jenslonkowski: What about also setting `current` and `previous` to undefined?

src/framework/services/scroll/ScrollPosition.js
```diff
@@ -0,0 +1,65 @@
...
+
+      // Default to the browser scroll position. Compare between values that are
+      // used in Chrome and Firefox, respectively.
+      return Math.max(
+        document.body.scrollTop,
+        document.documentElement.scrollTop
+      );
+    }
+
+    const getDistanceFromBottom = () => {
```
jenslonkowski: Why are you using different ways of defining `getPosition` and `getDistanceFromBottom`? I just checked the Airbnb style guide and they recommend using function declarations since they are named (https://github.com/airbnb/javascript#functions).

src/framework/services/scroll/ScrollPosition.js
```diff
@@ -0,0 +1,65 @@
+
+import ThrottledEventDispatcher from '../event/ThrottledEventDispatcher';
+
+export default class ScrollPosition {
+
+  constructor() {
+    this.listeners = [];
+    this.previous = this.current = 0;
```
jenslonkowski: I'd suggest to also set a default value for `this.distanceFromBottom` here.

src/framework/grid/Grid.spec.js
cjcenizal: Unfortunately, `create` can only substitute for `createFromFunction` or `createFromClass`, but not `createFromElement`.

jenslonkowski: I see, thanks.

src/framework/recycledList/RecycledList.jsx
cjcenizal: Length comparison may miss a substitution of items that are different, yet have the same length (which is far-fetched, but could result in really frustrating bugs).

In terms of whether it's working, we just need to pass in a fresh array in the render method  of the owner component to make sure this check works. This has been pretty standard practice in our code so far, but how about I add a comment to PropTypes to make this a little more explicit?

src/framework/recycledList/RecycledList.jsx
cjcenizal: It evaluates to false when `onScroll` is called in response to a scroll event, which delegates to `updateItemPositions`, which calls `setState`. So the state changes, but the props are the same.

jenslonkowski:
Great work CJ! I think these changes really improve the readability and of course the reusability.

I have a couple of comments and in Firefox (with Firebug activated) I can notice an increased lag while scrolling and I'm not sure where this is coming from - I think we have to investigate this (maybe in another PR).

src/framework/recycledList/RecycledList.jsx
cjcenizal: RecycledList receives a `scrollPosition` instance; it doesn't instantiate it itself. So since it doesn't own this instance, it seems like a bit of a weird side-effect for it to call `teardown`. But you bring up a good point that since it added a listener it should clean up after itself. I'll add a `removeListener` method which it can call in `componentWillUnmount`.

src/framework/services/scroll/ScrollPosition.js
cjcenizal: Hm, what's the benefit? I'm thinking of `teardown` as a method for avoiding memory leaks (by removing references to other objects), not really like a "reset".

src/framework/services/scroll/ScrollPosition.js
cjcenizal: I would have preferred a function declaration here too, but I'm referencing `this` in it, and fat arrow functions automatically bind to `this`.

src/framework/services/scroll/ScrollPosition.js
cjcenizal: What value should we use? We don't know the distance from the bottom because we don't have a reference to an element yet, and we can't say `0` because that would trigger a lazy-loading call in relevant situations.

jenslonkowski: That would be great. When I saw the examples I wondered where `distanceFromButton` was coming from. I thought maybe setting it to undefined in the constructor would make it clear that this is exposed on purpose.

ahollesen:
I think I understand most of the code ;) from what I get pretty awesome stuff. Buuut.. in firefox if you scroll down, then reload the page no elements are visible. One idea is to manually set the scroll position of elements to 0. As it seems chrome doesn't remember the element's scroll position but just the page (So MenuItem is no problem, Grid also not because its lazy loading). But this can be tackled in another ticket/PR. So as for now I think this is really great 👍

jenslonkowski:
Good catch Alan! If we have tickets for the performance and remembered scroll position I'm good to go as well! 👍

cjcenizal:
@jenslonkowski @ahollesen Good catch! I made a fix which partially addresses this problem (it fixed it for the Menu implementation), but I think there's still a problem with the Grid implementation. I've created these tickets to capture the problems you've found:

* https://jira.smaato.net/browse/BT-754
* https://jira.smaato.net/browse/BT-753

PR #236, opened 2016-06-03
[BT-764]: Added renderItem prop to RecycledList component

jenslonkowski:


src/framework/recycledList/RecycledList.spec.jsx
```diff
@@ -113,6 +115,24 @@ describe('RecycledList', () => {
         expect(props.itemHeightProvider).toHaveBeenCalled();
       });
     });
+
+    describe('renderItem', () => {
+      it('is used to render the items', () => {
```
ahollesen: Not really clear what is actually tested from the description. Perhaps just 'items', 'are rendered' or pass `renderItem` with default behaviour which would make the usage of the prop slightly more obvious!?

ahollesen:
One suggestion, up to you if you wanna change or not. Else looking good 👍

PR #240, opened 2016-06-16
[BT-790]: Added defaultValue prop to SearchBox...

jenslonkowski:
...and added reset method to ScrollPosition


ahollesen:
Looking good 👍

PR #242, opened 2016-06-28
[BT-748]: Added LOG type to Button component

jenslonkowski:


src/framework/button/_button.scss
```diff
@@ -24,6 +24,15 @@ $ctaButtonColor: #35d0a0;
     margin-left: 5px;
   }
 
+  .icon-history-blue {
+    height: 15px;
+    width: 15px;
+  }
```
ahollesen: What about adding `opacity: 1;`?

ahollesen:
Just one question, else good to go 👍

PR #250, opened 2016-07-14
Bt 841/only super users can change endpoints configuration

fedynapavlo:
Added readonly properties to BaseDropDown and TextInput

### DIFF
--- a/src/framework/baseDropdown/BaseDropdown.jsx
+++ b/src/framework/baseDropdown/BaseDropdown.jsx
@@ -196,7 +196,16 @@ export default class BaseDropdown extends Component {
         </div>
       );
     }
-
+    if (this.props.readOnly) {
+      return (
+        <div
+          data-id={this.props.dataId}
+          className={labelClasses}
+        >
+          {label}
+        </div>
+      );
+    }
     return (
       <div
         data-id={this.props.dataId}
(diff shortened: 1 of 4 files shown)


src/framework/textInput/TextInput.spec.js
```diff
@@ -64,5 +64,16 @@ describe('TextInput', () => {
         }
       );
     });
+
+    describe('readOnly', () => {
+      it('when input is readONly then readOnly attribute is applied', () => {
```
ahollesen: `readONly` typo. And what about the `it` description rather to be `adds the readOnly attribute`?

src/framework/baseDropdown/BaseDropdown.jsx
```diff
@@ -237,6 +246,7 @@ BaseDropdown.propTypes = {
   onSelect: PropTypes.func.isRequired,
   labelProvider: PropTypes.func.isRequired,
   optionLabelProvider: PropTypes.func.isRequired,
+  readOnly: PropTypes.bool,
```
ahollesen: Would be nice if prop definitions would be ordered alphabetically.

ahollesen:
Two small comments, else good to go 👍

jenslonkowski:
I made some improvements. From my side: 👍

fedynapavlo:
Last changes look fine.

PR #255, opened 2016-11-01
[SDX-1034] Create new Heading and Horizontal Line component

ahollesen:
- upgraded karma-phantomjs-launcher to make tests runnable again (sierraOS bug)
- adjusted failing test cases due to upgrade
- added size to heading, default is large (non-breaking)
- added small heading
- added horizontal line component

### DIFF
--- a/src/framework/framework.js
+++ b/src/framework/framework.js
@@ -39,6 +39,9 @@ export { default as Form } from './form/Form.jsx';
 export * from './grid/Grid.jsx';
 export { default as Grid } from './grid/Grid.jsx';
 
+export * from './horizontalLine/HorizontalLine.jsx';
+export { default as HorizontalLine } from './horizontalLine/HorizontalLine.jsx';
+
 export * from './kpi/Kpi.jsx';
 export { default as Kpi } from './kpi/Kpi.jsx';
 
--- a/src/framework/framework.js
+++ b/src/framework/framework.js
@@ -39,7 +39,6 @@ export { default as Form } from './form/Form.jsx';
 export * from './grid/Grid.jsx';
 export { default as Grid } from './grid/Grid.jsx';
 
-export * from './horizontalLine/HorizontalLine.jsx';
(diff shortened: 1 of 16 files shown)


src/framework/framework.js
```diff
@@ -39,6 +39,9 @@ export { default as Form } from './form/Form.jsx';
 export * from './grid/Grid.jsx';
 export { default as Grid } from './grid/Grid.jsx';
 
+export * from './horizontalLine/HorizontalLine.jsx';
```
jenslonkowski: This line is not needed because there are no other exports besides the default.

src/framework/text/Heading.spec.jsx
```diff
@@ -16,5 +16,29 @@ describe('Heading', () => {
...
     });
+
+    describe('size', () => {
+      describe('SMALL', () => {
+        it('adds the appropriate class', () => {
+          const props = {
+            size: Heading.SIZE.SMALL,
+          };
+
+          const testCase = TestCaseFactory.createFromFunction(Heading, props);
```
jenslonkowski: Let's use the "create" method for convenience.

src/framework/text/Heading.spec.jsx
```diff
@@ -16,5 +16,29 @@ describe('Heading', () => {
...
+          expect(testCase.dom.className).toContain('heading--small');
```
jenslonkowski: We could also use "toBe" to be more specific.

src/framework/text/Heading.spec.jsx
```diff
@@ -16,5 +16,29 @@ describe('Heading', () => {
...
+      });
+
+      describe('LARGE', () => {
+        it('adds the appropriate class', () => {
+          const props = {
+            size: Heading.SIZE.LARGE,
+          };
+
+          const testCase = TestCaseFactory.createFromFunction(Heading, props);
+          expect(testCase.dom.className).toContain('heading--large');
```
jenslonkowski: We could also use "toBe" to be more specific.

PR #258, opened 2016-11-07
[SDX-791] Added EscapeKeyHandler

jenslonkowski:


jenslonkowski:
Good catch @ahollesen - I removed `src/services.js`.

src/framework/services/event/EscapeKeyHandler.js
```diff
@@ -0,0 +1,23 @@
...
+    this.callback = callback;
+
+    this.onKeyDown = this.onKeyDown.bind(this);
+    this.removeListener = this.removeListener.bind(this);
+
+    document.addEventListener('keydown', this.onKeyDown);
+  }
+
+  onKeyDown(event) {
+    if (event.keyCode === 27) {
```
mohsinhijazee: If possible to introduce a constant here something like `ESCAPE_KEY`

src/framework/services/event/EscapeKeyHandler.spec.js
```diff
@@ -0,0 +1,61 @@
+
+import EscapeKeyHandler from './EscapeKeyHandler';
+
+describe('EscapeKeyHandler', () => {
+  describe('constructor', () => {
+    it('sets callback of instance to passed method', () => {
+      const callback = () => undefined;
```
mohsinhijazee: Could be `()=>{}`

jenslonkowski: Our code style doesn't allow this (http://eslint.org/docs/rules/no-empty-function).

PR #259, opened 2016-11-10
[SDX-1041] Create box component

ahollesen:
- moved BaseDropdown to general base package
- added BaseBox and Box
- added BodyMaxWidthLayout
- using BaseBox inside BodyPanel


src/framework/box/Box.jsx
```diff
@@ -0,0 +1,26 @@
+
+import classNames from 'classnames';
```
mohsinhijazee: Probably import from `classnames/dedupe` to avoid class name duplicates.

src/framework/box/Box.spec.jsx
```diff
@@ -0,0 +1,25 @@
...
+        const testCase = TestCaseFactory.create(Box, props);
+        expect(testCase.dom.className).toContain(props.classes);
+      });
+
+      it('don\'t replace "box" class', () => {
+        const props = {
+          classes: 'testClass',
+        };
+        const testCase = TestCaseFactory.create(Box, props);
+        expect(testCase.dom.className).toContain('box');
```
mohsinhijazee: Probably add a test for duplicate CSS classes if makes sense. Optional.

src/framework/box/Box.spec.jsx
```diff
@@ -0,0 +1,25 @@
...
+    describe('classes', () => {
+      it('are added to the element', () => {
+        const props = {
+          classes: 'testClass',
+        };
+        const testCase = TestCaseFactory.create(Box, props);
+        expect(testCase.dom.className).toContain(props.classes);
+      });
+
+      it('don\'t replace "box" class', () => {
```
mohsinhijazee: Replace `don't` with `doesn't`

src/framework/base/box/BaseBox.jsx
```diff
@@ -0,0 +1,30 @@
+
+import classNames from 'classnames/dedupe';
```
jenslonkowski: Why do we need dedupe here?

src/framework/base/box/BaseBox.spec.jsx
```diff
@@ -0,0 +1,57 @@
...
+        };
+        const testCase = TestCaseFactory.create(BaseBox, props);
+        expect(testCase.dom.getAttribute('data-id')).toBe(props.dataId);
+      });
+    });
+
+    describe('roundedCorners', () => {
+      it('when false doesn\'t add \'roundedCorners\' class', () => {
+        const props = {
+          roundedCorners: false,
```
jenslonkowski: The required prop `classes` is missing.

src/framework/base/box/BaseBox.spec.jsx
```diff
@@ -0,0 +1,57 @@
...
+          classes: '',
+          dataId: 'dataId',
+        };
+        const testCase = TestCaseFactory.create(BaseBox, props);
+        expect(testCase.dom.getAttribute('data-id')).toBe(props.dataId);
+      });
+    });
+
+    describe('roundedCorners', () => {
+      it('when false doesn\'t add \'roundedCorners\' class', () => {
```
jenslonkowski: `roundedCorners` can be in double quotation marks.

src/framework/base/box/BaseBox.spec.jsx
```diff
@@ -0,0 +1,57 @@
...
+        const props = {
+          roundedCorners: false,
+        };
+        const testCase = TestCaseFactory.create(BaseBox, props);
+        expect(testCase.dom.className).not.toContain('roundedCorners');
+      });
+
+      it('when true adds \'roundedCorners\' class', () => {
```
jenslonkowski: `roundedCorners` can be in double quotation marks.

src/framework/base/box/BaseBox.spec.jsx
```diff
@@ -0,0 +1,57 @@
...
+        const props = {
+          roundedCorners: true,
```
jenslonkowski: The required prop `classes` is missing.

src/framework/body/BodyPanel.jsx
```diff
@@ -1,18 +1,20 @@
...
   });
 
   return (
-    <div className={classes}>
+    <BaseBox classes={classes}>
       {props.children}
-    </div>
+    </BaseBox>
   );
 };
```
jenslonkowski: With this implementation it won't be possible to have a BodyPanel with rounded corners as we intended. BodyPanel needs to accept a prop `roundedCorners` and pass it to BasBox.

src/framework/box/Box.spec.jsx
```diff
@@ -0,0 +1,33 @@
...
+    describe('classes', () => {
+      it('are added to the element', () => {
+        const props = {
+          classes: 'testClass',
+        };
+        const testCase = TestCaseFactory.create(Box, props);
+        expect(testCase.dom.className).toContain(props.classes);
+      });
+
+      it('doesn\'t replace \'box\' class', () => {
```
jenslonkowski: `box` can be in double quotation marks.

src/framework/base/box/BaseBox.jsx
mohsinhijazee: Not sure, but removed. Test passes without it so not needed.

src/framework/box/Box.spec.jsx
mohsinhijazee: Plus, we can use _does not_ instead of _doesn't_ to avoid another single quote everywhere else in the same PR.

PR #261, opened 2016-11-14
[SDX-1044] Renamed Chart to LineChart and adjusted styling and rendering logic

jenslonkowski:


src/framework/chart/_lineChart.scss
```diff
@@ -0,0 +1,32 @@
...
+// Y Axis
+
+.lineChartYAxisTick__mark {
+  shape-rendering: crispEdges;
+  stroke: $lineChartGridColor;
+}
+
+.lineChartYAxisTick__text {
+  fill: $lineChartLabelColor;
+  font-size: 11px;
```
mohsinhijazee: The label colour, font size and tick colour are not configurable via props as far as it seems from the code. Neither they differ so in such case, probably we can have only two of the above four CSS classes for the same effect. 
But if it required for making the code more testable or something, then it is fine.

PR #263, opened 2016-11-16
Corrected framework exports and test file extensions

jenslonkowski:


src/framework/framework.js
```diff
@@ -17,21 +17,18 @@ export { default as ButtonGroup } from './buttonGroup/ButtonGroup.jsx';
...
 export { default as LineChart } from './chart/LineChart.jsx';
 
-export * from './checkBox/CheckBox.jsx';
 export { default as CheckBox } from './checkBox/CheckBox.jsx';
 
 export * from './columnLayout/ColumnLayout.jsx';
 export { default as ColumnLayout } from './columnLayout/ColumnLayout.jsx';
 
-export * from './dateRange/DateRange.jsx';
 export { default as DateRange } from './dateRange/DateRange.jsx';
```
ahollesen: I even think `DateRange` should not be exposed as it's not a framework component (at least yet) but only used in the `GridViewExample`.

PR #264, opened 2016-11-17
[SDX-1035] Added Chart component with example

jenslonkowski:


src/framework/chart/Chart.jsx
```diff
@@ -0,0 +1,121 @@
...
+} from './LineChart.jsx';
+
+export default class Chart extends Component {
+
+  constructor(props) {
+    super(props);
+
+    this.COLORS = ['#2799C4', '#35D0A0'];
+    this.HEIGHT = 520;
+
```
mohsinhijazee: Put in constants only if possible. Optional.

src/framework/chart/Chart.jsx
```diff
@@ -0,0 +1,121 @@
...
+          yAxisRange={[this.minY, this.maxY]}
+        />
+        {progress}
+      </div>
+    );
+  }
+
+  render() {
+    this.setData();
+
```
mohsinhijazee: This seems bit out of place. May be `componentWillUpdate()` or constructor can be used for that. So it is called in constructor, this will take care of the initial render. For subsequent renders, I think `componentWillUpdate()` would be able to compute. As per convention, `render()` would stay clean. Just a suggestion.

src/framework/chart/Chart.spec.jsx
```diff
@@ -0,0 +1,124 @@
...
+
+        expect(maxY).toBe(props.data[1][0].value);
+      });
+    });
+
+    describe('isLoading', () => {
+      const progressSelector = '.progress';
+
+      it('when false progress isn\'t rendered', () => {
+        const props = Object.assign({}, requiredProps, {
```
mohsinhijazee: `is not` can avoid `isn\'t`

PR #265, opened 2016-11-24
[SDX-1026] Chart and LineChart improvements

jenslonkowski:


package.json
```diff
@@ -15,7 +15,7 @@
   "dependencies": {
     "babel-core": "5.8.22",
     "classnames": "2.1.3",
-    "d3": "3.5.6",
+    "d3": "3.5.17",
```
mohsinhijazee: Intentional upgrade?

src/framework/services/number/Number.spec.js
```diff
@@ -0,0 +1,26 @@
+
+import Number from './Number';
+
+describe('Number', () => {
+  describe('formatBigNumber', () => {
+    it('doesn\'t format a number smaller than 1,000', () => {
+      expect(Number.formatBigNumber(123)).toEqual('123');
```
mohsinhijazee: What happens to the negative numbers? Probably we can add some tests here.

package.json
ahollesen: yes

PR #267, opened 2016-11-29
[SDX-1069] Added tooltip support to Chart and LineChart

jenslonkowski:


src/framework/chart/LineChart.jsx
```diff
@@ -163,6 +168,66 @@ export default class LineChart extends Component {
     // Transition each line from the previous shape to the new shape.
     lines.transition().duration(duration)
       .attr('d', item => lineGenerator(item.values));
+
+    // TOOLTIP /////////////////////////////////////////////////////////////////
```
mohsinhijazee: This marker indicates that probably the following code deserve to be a utility function or something for better organisation of the code.

src/framework/chart/_chart.scss
```diff
@@ -19,7 +19,6 @@
 
   .chart__lineChart {
     margin-top: 20px;
-    position: relative;
```
ahollesen: Should be added again for the `Progress`.

src/framework/chart/LineChart.jsx
jenslonkowski: In the context of the renderChart method it doesn't make sense to only extract that tooltip part but we might want to think of extracting parts of the method in general.

PR #268, opened 2016-12-07
[SDX-793] Added timeout to SearchBox for immediate searching

jenslonkowski:


src/framework/searchBox/SearchBox.spec.jsx
```diff
@@ -35,34 +48,43 @@ describe('SearchBox', () => {
...
-          expect(props.onSearch).not.toHaveBeenCalled();
+          setTimeout(() => {
+            expect(props.onSearch).not.toHaveBeenCalled();
+            done();
+          }, props.timeout);
         }
       );
 
       it(
         'is called when the user hits non-Enter and isImmediate is true',
```
ahollesen: You could consider adding another `describe` for `isImmediate`, but up to you

PR #271, opened 2017-01-10
[SDX-1097] LineChart improvements

jenslonkowski:
1. Y axis can now also be on the left.
2. Line dots closest to mouse cursor are always shown.

### DIFF
--- a/src/framework/chart/LineChart.jsx
+++ b/src/framework/chart/LineChart.jsx
@@ -16,16 +16,14 @@ export default class LineChart extends Component {
     super(props);
 
     this.DOT_RADIUS = 5;
+    this.DOT_SELECTOR = '.lineChartDot';
 
+    this.onMouseleave = this.onMouseleave.bind(this);
+    this.onMousemove = this.onMousemove.bind(this);
     this.onResize = this.onResize.bind(this);
   }
 
   componentDidMount() {
-    // Throttle resize event handling, in an attempt to improve performance.
-    this.resizeEventDispatcher = new ThrottledEventDispatcher(
-      'resize', `lineChartResize${new Date()}`, window, this.onResize
-    );
-
     this.$lineChart = $(this.refs.lineChart);
 
     this.lineChartSvg = d3.select(this.refs.lineChartSvg);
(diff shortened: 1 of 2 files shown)


src/framework/chart/LineChart.jsx
```diff
@@ -82,10 +119,20 @@ export default class LineChart extends Component {
...
     const marginBottom = 14;
-    const marginRight = yAxisLabelWidth;
+    const marginTop = 5;
     const width = this.$lineChart.width();
-    const availableWidth = width - marginRight;
-    const availableHeight = height - marginBottom;
+    const yAxisMax = height - marginBottom;
+    const yAxisTickSize = width - yAxisLabelWidth;
+
+    let marginLeft = 5;
```
mohsinhijazee: For it to be flexible as part of a generic UI toolkit, it would be a good (and reasonably small) investment to move  `marginTop`, `marginBottom` as props whereas `marginHorizontal` can be something that can be `marginLeft` or `marginRight` depending what value `YAxisLeft` has.

src/framework/chart/LineChart.jsx
```diff
@@ -82,10 +119,20 @@ export default class LineChart extends Component {
...
+    let marginRight = yAxisLabelWidth;
+
+    if (props.isYAxisLeft) {
+      marginLeft = yAxisLabelWidth;
+      marginRight = 5;
+    }
```
mohsinhijazee: Same as above.

src/framework/chart/LineChart.jsx
```diff
@@ -36,6 +37,42 @@ export default class LineChart extends Component {
     this.container = this.lineChartSvg.append('g');
 
     this.renderChart(this.props, true);
+
+    this.lineChartSvg.on('mousemove', () => {
+      if (d3.event.target.isSameNode(this.refs.lineChartSvg)) {
```
jenslonkowski: Good catch. Then this attempt of mine apparently doesn't work as intended. The problem is that if I remove the if and you move the mouse over a horizontal line the x position will be taken from that element which leads to the first two dots being displayed.

So I will need to change the way I get the x position of the mouse.

src/framework/chart/LineChart.jsx
jenslonkowski: I added `xAxisLabelHeight` since this can be influenced from the outside (like yAxisLabelWidth) but the margins just make sure that there is no overflow that is hidden. The space around the chart can and should be controlled by the surrounding component.

PR #275, opened 2017-01-26
[SDX-1109]: Updated React Date Range dependency and improved default …

jenslonkowski:
…props of Date Range

### DIFF
--- a/src/framework/viewHeader/dateRange/DateRange.jsx
+++ b/src/framework/viewHeader/dateRange/DateRange.jsx
@@ -155,7 +155,9 @@ DateRange.propTypes = {
   startDate: PropTypes.object,
 };
 
+const endDate = moment.utc();
+
 DateRange.defaultProps = {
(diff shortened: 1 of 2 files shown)


src/framework/viewHeader/dateRange/DateRange.jsx
```diff
@@ -155,7 +155,9 @@ DateRange.propTypes = {
   startDate: PropTypes.object,
```
mohsinhijazee: As you already would be aware, the tests for the `DateRange` are missing such as if the dates are being set properly, if the state is changing with the events and if the component is being rendered properly with the supposed structure and classes within the DOM. 

This would be a lot and I'd say this PR is okay to go without given the circumstances, but I'd like to do the tests for this component in a separate branch under your guidance (assuring you that it would be minimal) next week or as time permits. Or may be if there is a particular reason we can skip the tests for this, then we can leave as is.

PR #277, opened 2017-02-08
[SDX-1153] Update react and lint dependecies

ahollesen:
- updated versions
- adjust to new listing rules


.eslintrc
```diff
@@ -1,3 +1,25 @@
...
+      "imports": "always-multiline",
+      "exports": "always-multiline",
+      "functions": "ignore",
+    }],
+    "import/extensions": 0,
+    "import/no-extraneous-dependencies": 0,
+    "jsx-a11y/no-static-element-interactions": 0,
+    "react/forbid-prop-types": 0,
+    "react/jsx-filename-extension": 0,
+    "react/no-array-index-key": 0,
```
mohsinhijazee: I am thinking if we should enable this. Now? probably not but later may be.

ahollesen: So as discussed, I had to find a reasonable way of either fixing or adding the rule to the exclusions. The smaato own configs aren't really changing any rules, instead we can clean up over time, so all those exclusions will be fixed in future anyway.

PR #278, opened 2017-02-08
[SDX-1099] Create toggle button component

ahollesen:
- added react-toggle as dependency
- added component and adjusted styling


olga-andreeva:
@mohsinhijazee task/SDX-1153_update_versions should already be upgraded. For me tests didn't fail, can you check again?

muthan:
for me they also fail same error only on Linux

PR #285, opened 2017-02-24
SDX-1146 Adjust gridfilter design for Status

olga-andreeva:


mohsinhijazee:
Some suggestions in #287
1. All the forms are moved into single folder
2. ConditionCheckerForm is now SingleSelectionForm, stylesheets and styles are renamed accordingly
3. ConditionCheckerList is now called SelectedFilterList

Recommendations
1. SelectedFilterList should have a separate component SelectedFilter instead of rendering it in place.
2. SingleSelectedFilter should not use ref rather should rely on state.

PR #286, opened 2017-02-24
[SDX-1135] Design dropdown options

ahollesen:
- adjusted styling of dropdown list


src/guide/views/dropdown/DropdownExample.jsx
```diff
@@ -98,7 +98,7 @@ export default class DropdownExample extends Component {
 
         <Example title="Borderless with dot">
           <Dropdown
-            dotColor={DropdownDot.COLOR.BLUE}
+            dotColor={DropdownDot.COLOR.GREEN}
```
muthan: not that I min but what was wrong with BLUE ?

PR #290, opened 2017-03-06
SDX-1196 Adjust close modal icon

olga-andreeva:


olga-andreeva:
Forgot that it used the css plus icon (it used the symbol "+" before). Fixed.

PR #291, opened 2017-03-08
SDX-1201 Prevent filter duplication

olga-andreeva:


src/framework/filterControl/FilterControl.spec.jsx
```diff
@@ -55,4 +56,34 @@ describe('FilterControl', () => {
       });
     });
   });
+
+  describe('add button', () => {
```
olga-andreeva: I think it is clearer to have two wrapping `describe` blocks – for props and for the add button. The first one tests that the props are used correctly, the second one tests that the UI behaves as expected.

src/framework/filterControl/filterSelections/SelectedFilterList.spec.jsx
```diff
@@ -23,6 +24,42 @@ describe('SelectedFilterList', () => {
       });
     });
 
+    describe('render', () => {
```
olga-andreeva: I would move it outside `describe('props')`.

PR #295, opened 2017-03-30
[SDX-1174] Handle error case when druid not available

ahollesen:
- added alert component + example

### DIFF
--- a/src/framework/fieldMessage/_index.scss
+++ b/src/framework/fieldMessage/_index.scss
@@ -1,6 +1,6 @@
 
 $fieldMessage_borderColor: #e6e6ae;
-$fieldMessage_bgColor: #ffffc5;
+$fieldMessage_bgColor: $yellow;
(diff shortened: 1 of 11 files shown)


src/guide/views/notificationBar/AlertExample.jsx
```diff
@@ -0,0 +1,44 @@
+
```
muthan: I can't find this in the framework

src/framework/fieldMessage/_index.scss
```diff
@@ -1,6 +1,6 @@
 
 $fieldMessage_borderColor: #e6e6ae;
```
mariacasancho: should we add also this colour to the constants?

ahollesen: Agree, created a ticket for cleaning up all colors: https://jira.smaato.net/browse/SDX-1221, so out of scope.

PR #296, opened 2017-04-03
Feature/sdx 1147 integrate status filter for line item

mariacasancho:


src/framework/filterControl/forms/MultipleSelectFilterForm.spec.jsx
```diff
@@ -4,10 +4,15 @@ import MultipleSelectFilterForm from './MultipleSelectFilterForm.jsx';
 import {
   Filter,
   FilterOption,
+  OneOfOption,
 } from '../../services';
 
-describe('MultipleSelectFilterForm', () => {
-  const options = ['Active', 'Archived', 'Stopped'];
+fdescribe('MultipleSelectFilterForm', () => {
```
olga-andreeva: `fdescribe`. Then `'returns joined string with ONE_OF comparisonType'` test fails in `Filter.spec.js` (fixed by initializing `comparisonValue` to `[new OneOfOption(5), new OneOfOption(6)]`).

src/framework/filterControl/forms/MultipleSelectFilterForm.spec.jsx
```diff
@@ -76,12 +81,37 @@ describe('MultipleSelectFilterForm', () => {
...
           expect(testCase.find(
-            `.filterForm--multiSelect__checkbox:contains(${option})`).length
-          ).toBe(1);
+            `.filterForm--multiSelect__checkbox:contains(${option.value})`
+          ).length).toBe(1);
         });
       });
     });
 
+    it('options has different value and label ', () => {
```
olga-andreeva: I would not have an `it` block, but rather define value and label to be different by default (in lines 11-15). We can have a small test for `OneOfOption` that checks label when only value is provided.

src/framework/filterControl/forms/MultipleSelectFilterForm.spec.jsx
```diff
@@ -137,8 +167,10 @@ describe('MultipleSelectFilterForm', () => {
       const filter = props.onAddFilter.calls.argsFor(0)[0];
 
       expect(filter.comparisonValue.length).toBe(2);
-      expect(filter.comparisonValue.indexOf('Active')).not.toBe(-1);
-      expect(filter.comparisonValue.indexOf('Archived')).not.toBe(-1);
+      expect(filter.comparisonValue.filter(e => e.name === 'Active').length
+        ).not.toBe(-1);
```
olga-andreeva: Better:
```
expect(filter.comparisonValue.filter(
  e => e.value === 'Active').length).toBe(1);
```
Length will always be at least 0. Previously we checked that the element was present, now we need to check that there is one element satisfying the filter condition.
And it should be `e.value` instead of `e.name`.

src/framework/services/filter/Filter.js
```diff
@@ -42,7 +42,9 @@ export default class Filter {
         return index !== -1;
       }
       case ComparisonTypes.ONE_OF: {
-        return this.comparisonValue.indexOf(itemValue) !== -1;
+        const comparisonTypeValue =
```
olga-andreeva: `comparisonTypeValue` does not seem right. Better to use `normalizedComparisonValue` as in other cases.

src/framework/filterControl/forms/MultipleSelectFilterForm.spec.jsx
ahollesen: I like.

src/framework/filterControl/forms/MultipleSelectFilterForm.spec.jsx
ahollesen: What about checking the object rather than checking the content of it. Something like:
```
expect(filter.comparisonValue.length).toBe(2);
filter.comparisonValue.forEach((comparisonValueItem, index) => {
  expect(
    filter.comparisonValue[index]
  ).toBe(filterOption.comparisonParameters.oneOfOptions[index]);
});
```

And one empty like before the `it`. (Plus opening bracket on same level as closing)

src/framework/services/filter/Filter.js
ahollesen: Even better `normalizedComparisonValues` as it is an array.

src/guide/views/grid/gridExampleFilterOptions.js
```diff
@@ -2,6 +2,7 @@
 import {
   FilterOption,
   ComparisonTypes,
```
ahollesen: Alphabetic order.

src/framework/filterControl/FilterControl.jsx
```diff
@@ -183,7 +183,7 @@ export default class FilterControl extends Component {
 
 FilterControl.propTypes = {
   filterOptions: FilterOptionList.propTypes.filterOptions,
-  onAddFilter: PropTypes.func.isRequired,
+  onAddFilter: PropTypes.func,
   onRemoveSelectedFilter: FilterItem.propTypes.onRemoveSelectedFilter,
   onReplaceFilter: PropTypes.func.isRequired,
```
olga-andreeva: Maybe `onReplaceFilter` should not be required as well? Since it is not needed in case of only one irremovable filter. 
`onAddFilter` is still required in the form classes.

src/framework/services/filter/FilterOption.js
```diff
@@ -6,6 +6,8 @@ export default class FilterOption {
     this.comparisonType = config.comparisonType;
     this.getValue = config.getValue;
     this.name = config.name;
+    // eslint-disable-next-line
+    this.isRemovable = config.isRemovable === false ? false : true;
```
olga-andreeva: This line can be replaced with `this.isRemovable = config.isRemovable !== false;`, but then it will be less clear. I'm fine with how it is now.

src/framework/filterControl/forms/MultipleSelectFilterForm.jsx
```diff
@@ -19,16 +19,17 @@ export default class MultipleSelectFilterForm extends Component {
...
       selectedOptions = (new Array(this.options.length)).fill(false);
     } else {
-      selectedOptions = this.options.map(option => (
-        this.props.comparisonValue.indexOf(option) !== -1
-      ));
+      selectedOptions = this.options.map(option =>
+        props.comparisonValue.reduce(
+          (previous, element) =>
+            element.value === option.value || previous, false)
+      );
```
ahollesen: The line wrap is a little off here, remember opening and closing bracket on the same level, then if multilines try to wrap each parameter. So something like 
```
props.comparisonValue.reduce(
  (foundValue, element) => element.value === option.value || foundValue,
  false
)
```

Plus the reduce implementation is neat but a find would also do, as you look for a specific object. So:
```
selectedOptions = this.options.map(option =>
  Boolean(props.comparisonValue.find(
    element => element.value === option.value)
  )
);
```

src/framework/services/filter/Filter.spec.js
```diff
@@ -2,6 +2,10 @@
 import Filter from './Filter';
 import ComparisonTypes from './ComparisonTypes';
 
+import {
+  OneOfOption,
+} from '../../services';
```
ahollesen: The other imports are also being done directly so this could do as well.

src/framework/services/filter/FilterOption.js
ahollesen: Better put into same line like `this.isRemovable = config.isRemovable === false ? false : true; // eslint-disable-line no-unneeded-ternary`. 
But I actually like Olgas suggestion as well.. up to you.

src/framework/services/filter/OneOfOption.js
```diff
@@ -0,0 +1,9 @@
+
+export default class OneOfOption {
+
+  constructor(value, label) {
+    this.value = value;
+    this.label = label || value;
+  }
+
+}
```
ahollesen: You could write a short test to make sure that `label === value` if non was given in the constructor, but not sure if really needed.

src/framework/filterControl/forms/MultipleSelectFilterForm.jsx
```diff
@@ -19,16 +19,17 @@ export default class MultipleSelectFilterForm extends Component {
...
       selectedOptions = (new Array(this.options.length)).fill(false);
     } else {
-      selectedOptions = this.options.map(option => (
-        this.props.comparisonValue.indexOf(option) !== -1
-      ));
+      selectedOptions = this.options.map(option =>
+        Boolean(props.comparisonValue.find(
+          element => element.value === option.value)
+        )
+      );
```
olga-andreeva: I suggest to check if `find` returns `undefined` instead. `option.value` may equal to 0 and then `Boolean(find..)` will be always `false`.

src/framework/services/filter/OneOfOption.spec.js
```diff
@@ -0,0 +1,16 @@
+
+import OneOfOption from './OneOfOption';
+
+describe('OneOfOption', () => {
+  describe('interface', () => {
+    describe('constructor method', () => {
+      it(
+        'label is same as value when is not given',
+        () => {
+          const oneOfOption = new OneOfOption('active');
```
olga-andreeva: You can make 7-9 one line.

src/framework/filterControl/forms/MultipleSelectFilterForm.jsx
ahollesen: The `find` returns the current `element` or `undefined`, not `element.value`, so I'd think we are fine here.

olga-andreeva: @ahollesen you are right. I would still prefer comparing it with `undefined`, but I don't mind as it is now.

src/guide/views/grid/GridExample.jsx
```diff
@@ -63,7 +67,26 @@ const defaultState = {
...
+          oneOfOptions: [
+            new OneOfOption('Active'),
+            new OneOfOption('Stopped'),
+            new OneOfOption('Archived'),
+          ],
+        },
+      }),
+      [
+        new OneOfOption('Active'),
+        new OneOfOption('Stopped'),
```
ahollesen: Can reuse again here.

PR #303, opened 2017-05-05
[SDX-1256] No scroll when dropdown is opened v2

olga-andreeva:
* Choose dropdown's direction better

### DIFF
--- a/src/framework/base/dropdown/BaseDropdown.jsx
+++ b/src/framework/base/dropdown/BaseDropdown.jsx
@@ -48,16 +48,8 @@ export default class BaseDropdown extends Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
-    const $input = $(this.refs.input);
-    const $optionList = $(this.refs.optionList);
-    if (
-      !prevState.isOpen && this.state.isOpen &&
-      $input.offset().top + $optionList.height() > $(window).height()
-    ) {
-      $optionList.css({
-        top: 'inherit',
-        bottom: 'calc(100% - 1px)',
-      });
+    if (!prevState.isOpen && this.state.isOpen) {
+      this.chooseOpeningDirection();
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/framework/base/dropdown/BaseDropdown.jsx
```diff
@@ -164,17 +156,44 @@ export default class BaseDropdown extends Component {
...
+
+    const $optionList = $(this.refs.optionList);
+
+    const roomBelow =
+      (
+        $(window).height() - $input.offset().top - $input.parent().outerHeight()
+      ) + $(window).scrollTop();
+    const roomAbove = $input.offset().top - $(window).scrollTop();
+
+    const EXTRA_SPACE = 10;
```
mariacasancho: I think constants should be at the beginning

src/framework/base/dropdown/BaseDropdown.jsx
```diff
@@ -164,17 +156,44 @@ export default class BaseDropdown extends Component {
...
+
+    if (roomBelow < $optionList.outerHeight()) {
```
mariacasancho: you can also do this "if" after getting the const roomBelow (before doing roomAbove) and return as you did with input.offset() so you save calculating roomAbove incase of false

src/framework/base/dropdown/BaseDropdown.jsx
mohsinhijazee: Would be better to move it up.

olga-andreeva: I think this constant makes sense only in the context of `chooseOpeningDirection`. I don't know how to name it if it is in the constructor.

src/framework/base/dropdown/BaseDropdown.jsx
olga-andreeva: If I understand you correctly, I should move `roomAbove` inside the `if` block.

PR #304, opened 2017-05-10
Feature/sdx 1235 adding button icons

mariacasancho:


src/framework/button/_button.scss
```diff
@@ -12,6 +12,22 @@ $ctaButtonColor: #35d0a0;
     height: $iconSize;
     width: $iconSize;
     font-size: 10px;
+
+    &.icon-deal {
+      margin-right: 4px;
+      transform: scale(2.7);
+      backface-visibility: hidden;
```
olga-andreeva: Why is `backface-visibility` needed?

olga-andreeva:
Some discrepancies between the browsers:
* `ButtonGroup` is not displayed properly in Safari (it has extra space in between the buttons)
* Labels in `ButtonGroup` are not aligned vertically to the center in Firefox
* Background color of the default button is different in Firefox. We should set the color ourselves.

src/framework/button/_button.scss
mariacasancho: because backface-visibility simplifies the "over animation" to the front of that object. If you remove backfase-visibility the icon will be jumping everytime you over the other button due to the pseudo-element that is created when you over and the transform scale.

src/framework/button/_button.scss
```diff
@@ -12,6 +12,22 @@ $ctaButtonColor: #35d0a0;
     height: $iconSize;
     width: $iconSize;
     font-size: 10px;
+
+    &.icon-deal {
+      margin-right: 4px;
+      transform: scale(2.7);
```
mariacasancho: images are different :)

PR #305, opened 2017-06-07
[SDX-1265] Add DISABLED to StatusDropdown

ahollesen:
- added DISABLED as possible option
- preventing rendering the dropdown in case of DISABLED

### DIFF
--- a/src/framework/statusDropdown/StatusDropdown.jsx
+++ b/src/framework/statusDropdown/StatusDropdown.jsx
@@ -55,6 +55,13 @@ export default class StatusDropdown extends Component {
           inactive: 'Pause',
         },
       },
+      [StatusDropdown.OPTIONS.DISABLED]: {
+        dropdownDotColor: DropdownDot.COLOR.GREY,
+        labelClass: 'statusDropdownLabel--grey',
+        name: {
+          active: 'Disabled',
+        },
+      },
     };
   }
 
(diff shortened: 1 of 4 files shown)


src/framework/statusDropdown/StatusDropdown.jsx
```diff
@@ -114,7 +121,11 @@ export default class StatusDropdown extends Component {
       this.optionToMetaDataMap[this.props.selectedOption].labelClass;
     const labelClasses = classNames('statusDropdownLabel', {
       [additionalLabelClass]: additionalLabelClass,
+      'statusDropdownLabel--disabled':
+        this.props.selectedOption === StatusDropdown.OPTIONS.DISABLED,
     });
+    const isReadonly =
+      this.props.selectedOption === StatusDropdown.OPTIONS.DISABLED;
```
mariacasancho: we should declare this before labelClasses, so we can call isReadonly instead of doing the comparison again

src/framework/statusDropdown/_statusDropdown.scss
```diff
@@ -39,11 +43,11 @@
     background-color: $dropdownDotGreyLine;
   }
 
-  &:before {
+  &:not(&--disabled):before {
     transform: rotate(-45deg) translate(-2.5px, -2.5px);
   }
 
-  &:after {
+  &:not(&--disabled):after {
```
mariacasancho: &:not(&--disabled) {
  &:before, &:after {
  }
  &:before {
  }
 &:after {
  }
}

or you can also set the before and after ($dropdownArrow_width....) in a placeholder selector and extend it when you have the single selectors for before and after http://sass-lang.com/documentation/file.SASS_REFERENCE.html#placeholders

ahollesen: Did first approach.

PR #308, opened 2017-07-13
[SDX-1342] create tooltip ui components for audience targeting

mariacasancho:


src/framework/tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,134 @@
...
+
+    const windowLimits = {
+      height: window.innerHeight,
+      width: window.innerWidth,
+    };
+
+    ReactDOM.render(
+      this.renderTooltipLabel(item),
+      this.tooltip
+    );
```
olga-andreeva: What about having `isVisible` variable in the state of the component? It would be set to true on `mouseenter` and to false on `mouseleave`. Then the render method would check it and display the tooltip accordingly. In that way, all rendering will happen inside `render`.

src/framework/tooltip/Tooltip.spec.jsx
```diff
@@ -0,0 +1,18 @@
+
+import { TestCaseFactory } from 'react-test-kit';
+import Tooltip from './Tooltip.jsx';
+
+fdescribe('Tooltip', () => {
```
olga-andreeva: f should be removed in the beginning.

src/framework/tooltip/_tooltip.scss
```diff
@@ -0,0 +1,50 @@
...
+
+.tooltipPosition {
+  display: none;
+  opacity: 0;
+  position: absolute;
+
+  &:after {
+    background-color: $fieldMessage_bgColor;
+    border: solid $fieldMessage_borderColor;
+    border-radius: 3px;
```
olga-andreeva: I think `border-radius` is not needed here. With it, the corners of the arrow become rounded.

src/guide/views/tooltip/TooltipExample.jsx
```diff
@@ -0,0 +1,31 @@
...
+      <Tooltip
+        width="200px"
+        message="This is a text to display in the tooltip"
+      >
+      Hover me
+      </Tooltip>
+    </Example>
+
+  </Page>
+);
```
olga-andreeva: What about adding an example with the fixed height as well?

src/framework/tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,134 @@
...
+    this.renderTooltip(item);
+  }
+
+  onMouseLeave() {
+    this.tooltip.style.display = 'none';
+    this.tooltip.style.opacity = 1;
+  }
+
+  getTooltipDimension() {
+    const tooltip = document.getElementById('tooltip');
```
olga-andreeva: Why not using jquery as in, for example,`BaseDropdown`?

src/framework/tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,134 @@
...
+    let lx;
+    let ly;
+    let element = tooltip;
+    for (lx = 0, ly = 0;
+      element != null;
+      lx += element.offsetLeft,
+      ly += element.offsetTop,
+      element = element.offsetParent);
+
+    return {
```
olga-andreeva: At least for the example, these equalities are true:
```
ly = $(element).offset().top
lx = $(element).offset().left
```

src/framework/tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,134 @@
...
+
+    const tooltipDimension = this.getTooltipDimension();
+    let tooltipX = (document.getElementById('tooltipParent').clientWidth / 2) -
+      (tooltipDimension.width - 15);
+    if (tooltipDimension.posX - tooltipDimension.width < 0) {
+      tooltipX = 0;
+      styles += ' tooltipLeft';
+    }
+
+    let tooltipY = document.getElementById('tooltipParent').clientHeight;
```
olga-andreeva: `$(this.refs.tooltipParent).height()` seems to have the same result.

src/framework/tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,134 @@
...
+        className="tooltipParent"
+        ref="tooltipParent"
+        id="tooltipParent"
+        onMouseEnter={this.onMouseEnter}
+        onMouseLeave={this.onMouseLeave}
+      >
+        {this.children}
+        <div
+          id="tooltipPosition"
+          className="tooltipPosition "
```
olga-andreeva: Why is it called `tooltipPosition` if that holds the actual tooltip?

src/framework/tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,134 @@
...
+    super(props);
+
+    this.content = this.props.message;
+    this.children = this.props.children;
+    this.onMouseEnter = this.onMouseEnter.bind(this);
+    this.onMouseLeave = this.onMouseLeave.bind(this);
+  }
+
+  componentDidMount() {
+    this.tooltip = this.refs.tooltipPosition;
```
ahollesen: Little confusing that `this.tooltip` is actually not the element with ref `tooltip` but `tooltipPosition`.

src/framework/tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,134 @@
...
+      (tooltipDimension.width - 15);
+    if (tooltipDimension.posX - tooltipDimension.width < 0) {
+      tooltipX = 0;
+      styles += ' tooltipLeft';
+    }
+
+    let tooltipY = document.getElementById('tooltipParent').clientHeight;
+    if (
+      windowLimits.height < tooltipDimension.height +
+      (tooltipDimension.posY + 20)
```
ahollesen: The bracketing is weird. Is needed at all?

src/framework/tooltip/_tooltip.scss
```diff
@@ -0,0 +1,50 @@
...
+    border-width: 0 1px 1px 0;
+    top: initial;
+  }
+
+  &.tooltipLeft:after {
+    left: 20px;
+  }
+
+  .box {
+    box-shadow: 0 2px 4px 0 rgba(189, 189, 189, 0.5);
```
ahollesen: Slightly nicer to use `rgba(#bdbdbd, 0.5)` (or reuse another grey tone) and then perhaps also extract it to `_colors.scss` as some `$shadowColor`.

src/framework/tooltip/Tooltip.jsx
mariacasancho: I think this project shouldn't have jquery so I try to use vanilla javascript

src/framework/tooltip/Tooltip.jsx
mariacasancho: Everytime that you use $(element) it doesn't only get the element, it also converts it to a jquery object. It doesn't make sense to get the element, convert it to a jquery object and get the height with a function from that jquery object when you can just have the height from the element.

src/framework/tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,134 @@
...
+    );
+  }
+
+  renderTooltip(item) {
+    this.tooltip.style.display = 'inline-block';
+    this.tooltip.style.left = '0px';
+    this.tooltip.style.top = '0px';
+    this.tooltip.style.height = this.props.height;
+    this.tooltip.style.width = this.props.width;
+    let styles = 'tooltipPosition ';
```
mariacasancho: I don't think it is better to create an array and convert it to string while you just can have the string directly, we aren't replacing strings or segregating them or.., it is just concat so I don't see the danger. What it needs to be done will not get any benefit from doing it with array and it will just have less performance.

src/framework/tooltip/Tooltip.jsx
mariacasancho: the idea about using display: none vs visibility: hidden is that with display: none the element is ignored from the rendering, visibility: hidden, renders the element and occupies space (it is just not visible). 

So if you reload the page and you don't over, you will not see the element there with display: none. I think a tooltip will not be always be used so I think it is better to gain a little bit of performance while we load the page by using display: none.

src/framework/tooltip/Tooltip.jsx
olga-andreeva: Ok

src/framework/tooltip/_tooltip.scss
```diff
@@ -0,0 +1,50 @@
+
+$fieldMessage_arrowSide: 6px;
+$fieldMessage_bgColor: #fafdfe;
+$fieldMessage_borderColor: #dbe2e3;
```
mariacasancho: Background-color and border-color are taken from the mock ups from Nithin from https://projects.invisionapp.com/d/main#/console/11263450/239311720/inspect

src/framework/tooltip/_tooltip.scss
mariacasancho: colors are copied from mock up from Nithin https://projects.invisionapp.com/d/main#/console/11263450/239311720/inspect
I prefer that the extraction will be in the ticket reserved for it to choose correct naming for all colors used in UI framework.

src/framework/tooltip/Tooltip.jsx
olga-andreeva: I thought jQuery does more than that – it also takes care of differences between the browsers. But if it is not an issue anymore and we want to move away from jQuery – all good.

olga-andreeva:
I confirmed with Jens that the right way to get DOM elements in React is by using refs. The `id` should not be used and in our case it will result in an invalid HTML. So I created a PR that changes ids to refs – https://github.com/smaato/ui-framework/pull/309. It seems to work and there are no warnings in the console.

src/framework/tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,136 @@
...
+          </div>
+        </Box>
+      </div>
+    );
+  }
+
+  render() {
+    let tooltipContainter;
+
+    if (this.state && this.state.hover) {
```
olga-andreeva: Do we need to check `this.state`? Since it is created in the constructor, it must always exist.

src/framework/tooltip/_tooltip.scss
olga-andreeva: The background color of the arrow and of the box is different. The box is still white.

src/framework/tooltip/_tooltip.scss
```diff
@@ -0,0 +1,50 @@
...
+  &:after {
+    background-color: $bgColor;
+    border: solid $borderColor;
+    border-width: 1px 0 0 1px;
+    content: '';
+    display: block;
+    height: $arrowSide;
+    position: absolute;
+    margin-left: -$arrowSide;
+    margin-top: -4px;
```
olga-andreeva: At least on chrome, `margin-top` should be 3. Otherwise there is an overlap of borders.

src/framework/tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,136 @@
...
+  constructor(props) {
+    super(props);
+
+    this.onMouseEnter = this.onMouseEnter.bind(this);
+    this.onMouseLeave = this.onMouseLeave.bind(this);
+    this.state = { hover: false };
+  }
+
+  componentDidUpdate() {
+    if (this.state && this.state.hover) {
```
ahollesen: No need of checking `this.state` as defined in constructor.

src/framework/tooltip/Tooltip.jsx
ahollesen: The danger lies in the needed spaces. The benefit is that you don't have to take care of setting the spaces yourself.

src/framework/tooltip/_tooltip.scss
```diff
@@ -0,0 +1,50 @@
...
+
+  &.tooltipLeft:after {
+    left: 20px;
+  }
+
+  .box {
+    box-shadow: 0 2px 4px 0 rgba(189, 189, 189, 0.5);
+  }
+
+}
```
ahollesen: Indentation in css should mimic the html structure generated so something like 
```
.tooltipWrapper {
}

  .tooltipContainer { <-- selector not needed to be cascaded but indication of child relation
  }

    .tooltip { <-- same here: tooltip is being rendered inside the tooltipContainer
    }

File: src/guide/views/tooltip/TooltipExample.jsx [hunk 18]
```diff
@@ -0,0 +1,38 @@
...
+  <Page title={props.route.name}>
+
+    <Example>
+      <Tooltip
+        width="200px"
+        message="This is a text to display in the tooltip"
+      >
+      Hover me
+      </Tooltip>
+      <br />
```
ahollesen: It would be nice to split the examples into actual `Example`s.

src/framework/tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,135 @@
...
+
+      this.windowLimits = {
+        height: window.innerHeight,
+        width: window.innerWidth,
+      };
+
+      const tooltipDimension = this.getTooltipDimension();
+      let tooltipX = (
+          this.tooltipWrapper.clientWidth / 2
+        ) - (tooltipDimension.width - 15);
```
ahollesen: Line wrapping a little off, what about 
```
let tooltipX =
  (this.tooltipWrapper.clientWidth / 2) - (tooltipDimension.width - 15);
```

src/framework/tooltip/_tooltip.scss
```diff
@@ -0,0 +1,52 @@
+
+$arrowSide: 6px;
+$bgColor: $white;
+$borderColor: $componentStrokeColor;
+
+.tooltipWrapper {
+  display: inline-block;
+  position: relative;
+}
```
ahollesen: Needs a `line-height: 0;` to prevent calculating the position on wrong premises.

src/framework/tooltip/_tooltip.scss
```diff
@@ -0,0 +1,52 @@
...
+
+  .tooltipContainer {
+    display: inline-block;
+    opacity: 0;
```
ahollesen: Not needed.

src/framework/tooltip/_tooltip.scss
```diff
@@ -0,0 +1,51 @@
+
+$arrowSide: 6px;
+
+.tooltip {
+  display: inline-block;
+  position: relative;
+  line-height: 0;
```
ahollesen: Order.

src/framework/tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,128 @@
...
+        tooltipDimension.posY + 20
+      ) {
+        tooltipY = -tooltipDimension.height - 19;
+        this.styles.push('tooltipTop');
+      }
+
+      this.container.className = this.styles.join(' ');
+      this.container.style.left = `${tooltipX}px`;
+      this.container.style.top = `${tooltipY}px`;
+      this.container.style.opacity = 1;
```
ahollesen: Opacity not needed (as at least I don't see any flickering).

src/framework/tooltip/_tooltip.scss
```diff
@@ -0,0 +1,51 @@
...
+
+  &__container {
+    display: inline-block;
+    opacity: 0;
```
ahollesen: Should not be needed.

src/framework/tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,128 @@
...
+
+    this.onMouseEnter = this.onMouseEnter.bind(this);
+    this.onMouseLeave = this.onMouseLeave.bind(this);
+    this.state = { hover: false };
+  }
+
+  componentDidUpdate() {
+    if (this.state.hover) {
+      this.content.style.width = this.props.width;
+      this.styles = ['tooltip__container'];
```
ahollesen: Not needed  to store in `this`. `const styles =...` would do.

PR #310, opened 2017-07-31
[SDX-1389] update tooltip component in ui framework

mariacasancho:


src/framework/tooltip/Tooltip.jsx
```diff
@@ -43,11 +47,46 @@ export default class Tooltip extends Component {
...
         styles.push('tooltipTop');
+        isPosYSet = true;
       }
 
-      this.container.className = styles.join(' ');
       this.container.style.left = `${tooltipX}px`;
       this.container.style.top = `${tooltipY}px`;
+
+      // reajust tooltip taking into account if it is visible
+      const containerBounds = this.container.getBoundingClientRect();
```
mohsinhijazee: I would still consider moving this calculation to a utility method that returns heigh, width, top and left for the position for tooltip. Additionally, may be we can have a few unit tests for that utility method. Depends how easy it is, that would be up to you and @ahollesen if that's feasible or whether makes sense.

src/framework/tooltip/Tooltip.jsx
```diff
@@ -28,12 +28,16 @@ export default class Tooltip extends Component {
         width: window.innerWidth,
```
ahollesen: Not used anymore.

src/framework/tooltip/Tooltip.jsx
```diff
@@ -43,11 +47,46 @@ export default class Tooltip extends Component {
       ) {
         tooltipY = -tooltipDimension.height - 19;
         styles.push('tooltipTop');
+        isPosYSet = true;
       }
 
-      this.container.className = styles.join(' ');
       this.container.style.left = `${tooltipX}px`;
       this.container.style.top = `${tooltipY}px`;
+
```
ahollesen: You might wanna split it into two methods for better structuring and understanding why it has to calculated twice. Something like: 
```
componentDidUpdate() {
  if (this.state.hover) {
    calculatePositionToWindow()
    adjustPositionForVisibility()
  }
}
```

Also consider reusing calculations a little (e.g. see the updating the position to the right, basically `tooltipX`, setting style and class is the same).

PR #313, opened 2017-09-08
[SDX-1419] create action dropdown

mariacasancho:


src/framework/dropdown/DropdownGroup.jsx
```diff
@@ -0,0 +1,101 @@
...
+import DropdownDot from './dropdownDot/DropdownDot.jsx';
+import DropdownOption from './DropdownOption.jsx';
+
+export {
+  default as DropdownDot,
+} from './dropdownDot/DropdownDot.jsx';
+
+export {
+  default as DropdownOption,
+} from './DropdownOption.jsx';
```
ahollesen: Exports already happening in `Button`, no need for exporting (even thinking if this should somehow fail).

src/framework/dropdown/DropdownGroup.jsx
```diff
@@ -0,0 +1,101 @@
...
+    []);
+
+    const extendedProps = Object.assign({}, this.props, {
+      options: optiontolist,
+      labelClasses,
+      labelProvider: this.labelProvider,
+    });
+
+    return (
+      <BaseDropdown
```
ahollesen: Think it would make more sense to render a `Dropdown` here with altered `options` instead. Then we don't redo the thing.

src/framework/dropdown/DropdownGroup.jsx
```diff
@@ -0,0 +1,101 @@
...
+    return label;
+  }
+
+  render() {
+    const labelClasses = classNames(this.props.labelClasses, {
+      'dropdownLabel--borderless': this.props.isBorderless,
+      'dropdownLabel--readonly': this.props.isReadonly,
+    });
+
+    const optiontolist = this.props.options.reduce(
```
ahollesen: Camel case. Plus the result is the list itself (like `options` or `optionList`), but `optionToList` indicates a method rather than a list I think.

src/framework/dropdown/DropdownGroup.jsx
```diff
@@ -0,0 +1,101 @@
...
+}
+
+DropdownGroup.propTypes = Object.assign({}, BaseDropdown.propTypes, {
+  dotColor: DropdownDot.propTypes.color,
+  isBorderless: PropTypes.bool,
+  optionsClass: PropTypes.array,
+});
+
+DropdownGroup.defaultProps = Object.assign({}, BaseDropdown.defaultProps, {
+  classes: 'dropdown',
```
ahollesen: The default should be `dropdownGroup` rather than `dropdown`.

src/framework/dropdown/DropdownGroup.jsx
```diff
@@ -0,0 +1,41 @@
+
+import React, {
+  PropTypes,
+} from 'react';
+import classNames from 'classnames';
```
ahollesen: Order by lib name, so `'classnames'` before `'react'`.

src/framework/dropdown/DropdownGroup.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+import Dropdown from './Dropdown.jsx';
+
+const DropdownGroup = (props) => {
+  const optionToList = props.options.reduce(
```
ahollesen: Comment from initial review:

> The result is the list itself (like `options` or `optionList`), but optionToList indicates a method rather than a list I think.

src/framework/dropdown/DropdownGroup.jsx
```diff
@@ -0,0 +1,41 @@
...
+    (accumulator, optionList, currentIndex) => (
+      accumulator.concat(optionList.map((option, index) => (
+        Object.assign({}, option, {
+          class: classNames(props.optionsClass[currentIndex], {
```
ahollesen: What about calling this `classes` instead as that plural usually?

src/framework/dropdown/DropdownGroup.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+  return (
+    <Dropdown
+      {...extendedProps}
+    />
+  );
+};
+
+DropdownGroup.propTypes = Object.assign({}, Dropdown.propTypes, {
+  optionsClass: PropTypes.array,
```
ahollesen: What about `optionGroupClasses` or something?

src/framework/dropdown/DropdownGroup.jsx
```diff
@@ -0,0 +1,41 @@
...
+});
+
+DropdownGroup.defaultProps = Object.assign({}, Dropdown.defaultProps, {
+  classes: 'dropdownGroup',
+});
+
+export default DropdownGroup;
```
ahollesen: Still think that `DropdownGroup` could be misleading. It rather indicates that it's a group of dropdowns. Perhaps in direction of `GroupedOptionsDropdown`?! We have a `ButtonGroup` which is a group of buttons (specifically `GroupedButton`s which is perhaps also not perfect wording).

src/framework/dropdown/DropdownGroup.spec.jsx
```diff
@@ -0,0 +1,21 @@
...
+import { TestCaseFactory } from 'react-test-kit';
+
+import Dropdown from './Dropdown.jsx';
+import DropdownGroup from './DropdownGroup.jsx';
+
+describe('DropdownGroup', () => {
+  const requiredProps = {
+    options: [['blue', 'pink'], ['black', 'white']],
+    optionsClass: ['', 'dropdownClass'],
+    labelProvider: () => undefined,
```
ahollesen: Order.

src/framework/dropdown/DropdownGroup.spec.jsx
```diff
@@ -0,0 +1,21 @@
...
+  };
+
+  describe('DOM structure', () => {
+    it('is a Dropdown', () => {
+      const props = Object.assign({}, requiredProps);
+      const testCase = TestCaseFactory.create(DropdownGroup, props);
+      expect(testCase.findComponents(Dropdown)).toBeDefined();
```
ahollesen: `findComponents` return a list. `[]` will also be defined. So not rendering a `Dropdown` is not caught by the test.

src/framework/dropdown/DropdownOption.jsx
```diff
@@ -1,13 +1,21 @@
 
 import React from 'react';
 
+import classNames from 'classnames';
```
ahollesen: External imports should be grouped together, and then order.

src/framework/dropdown/_dropdownOption.scss
```diff
@@ -9,4 +9,8 @@
   &.is-dropdown-option-focus {
     background-color: #f3f9fd;
   }
+
+  &.dropdownGroupFirst {
```
ahollesen: BEM style would rather be `dropdownGroup--first` or even `dropdownGroup__dropdownOption--first`.

src/framework/dropdown/_dropdownOption.scss
```diff
@@ -9,4 +9,8 @@
...
+    border-top: 1px dashed #7a8183;
```
ahollesen: In the design there is an opacity of 40% which makes the dashed line lighter.

src/guide/views/dropdown/DropdownExample.jsx
```diff
@@ -76,6 +94,16 @@ export default class DropdownExample extends Component {
...
 
+        <Example title="With group">
+          <DropdownGroup
+            labelProvider={this.dropdownLabelProvider}
+            options={this.dropdownItemsGroup}
+            optionsClass={this.dropdownItemsClass}
+            onSelect={this.onSelectDropdownOption}
+            optionLabelProvider={this.dropdownOptionLabelProvider}
+          />
+        </Example>
```
ahollesen: Selecting an option doesn't change the label, but changes the content of the other dropdowns.

src/guide/views/dropdown/DropdownExample.jsx
```diff
@@ -48,6 +49,23 @@ export default class DropdownExample extends Component {
...
+    }, {
+      name: 'Cat',
+    }, {
+      name: 'Dog',
+    }, {
+      name: 'Elephant',
+    }];
+
+    this.dropdownItemsGroup =
+      [this.dropdownItems, this.dropdownItemsAnimals];
```
ahollesen: One liner?

src/framework/dropdown/DropdownGroup.jsx
mariacasancho: I don't think it is misleading because in the rare case that ui framework will need to provide a component that provides a group of dropdowns it will be name dropdowns as plural

src/framework/dropdown/_dropdownOption.scss
```diff
@@ -9,4 +9,8 @@
   &.is-dropdown-option-focus {
     background-color: #f3f9fd;
   }
+
+  &.dropdownGroup--first {
+    border-top: 1px dashed rgba(122, 129, 131, 0.4);
```
ahollesen: `1px dashed rgba(#7a8183, 0.4)` would also do. Or even a predefined color (as of not introducing too many different colors even though mockup does). Up to you...

src/framework/dropdown/DropdownGroup.jsx
```diff
@@ -0,0 +1,41 @@
...
+
+  return (
+    <Dropdown
+      {...extendedProps}
+    />
+  );
+};
+
+DropdownGroup.propTypes = Object.assign({}, Dropdown.propTypes, {
+  optionGroupClasses: PropTypes.array,
```
ahollesen: One more thought: This array is not required and should probably also not, which means `props.optionGroupClasses[currentIndex]` would break. As of our usage right now it doesn't matter because we need it anyway. But else more liberal handling would be nice.

src/framework/dropdown/DropdownGroup.jsx
```diff
@@ -0,0 +1,41 @@
+
+import classNames from 'classnames';
+import React, {
+  PropTypes,
+} from 'react';
+
+import Dropdown from './Dropdown.jsx';
+
+const DropdownGroup = (props) => {
```
mohsinhijazee: Possibly - `GroupedDropdown`

src/framework/dropdown/_dropdownOption.scss
mohsinhijazee: `z-index` should be adjusted so that the line is drawn below the scrollbar and not over it.

src/framework/dropdown/DropdownGroup.jsx
mohsinhijazee: Optionally, there's no harm if we allow people to define the class for separator as well when we already allow each group to have a different class. Up to you.

src/framework/dropdown/_dropdownOption.scss
mariacasancho: scrollbar is transparent

src/framework/dropdown/DropdownGroup.jsx
mariacasancho: I though about that, but it can also be controlled by the group class, in the example you can do 
```
.dropdownLabel--red.dropdownGroup--first {
    border-color: red;
}
```
to change the border color of the separator

PR #315, opened 2017-11-09
Updated git dependencies and URL to icons.css

jenslonkowski:


src/guide/index.jade
```diff
@@ -23,7 +23,7 @@ html(lang='en')
 
     //- Icons
     link(
-      href = "//smaatoui.github.io/icon-library/css/icons.css"
+      href = "//s3.amazonaws.com/smt-ui-icon-library/icons.css"
```
mohsinhijazee: Thinking if it should be hosted on smaato.github.io instead of S3.

package.json
```diff
@@ -46,19 +46,19 @@
...
+    "eslint-config-smaato": "git+https://github.com/smaato/eslint-config-smaato.git#v0.1.0",
     "eslint-plugin-import": "2.2.0",
     "eslint-plugin-jasmine": "1.6.0",
     "eslint-plugin-jsx-a11y": "4.0.0",
     "eslint-plugin-react": "6.9.0",
     "gulp": "3.9.0",
     "gulp-gh-pages": "0.5.4",
     "gulp-replace": "0.5.4",
-    "gulp-tasks": "git@github.com:Smaato/gulp-tasks.git#v0.7.0",
+    "gulp-tasks": "git+https://github.com/smaato/gulp-tasks.git#v0.7.0",
```
mohsinhijazee: This particular `gulp-tasks 0.7.0` still points to repository that's [only privately accessible](https://github.com/smaato/gulp-tasks/blob/v0.7.0/package.json#L49) and I do not know what's the resolution is. 

One way could be to do a 0.7.1 release for `gulp-tasks` that points to `eslint-config-smaato` using https instead of ssh and point current `ui-framework` to `gulp-tasks` 0.7.1

mohsinhijazee: Please ignore, I tried it inside a new node docker container and was able to install dependencies.

src/guide/index.jade
jenslonkowski: This is only a temporary solution.

PR #316, opened 2017-12-07
[SEM-26] Create list view for mapped external inventory:

mariacasancho:
- alert button implemented.

### DIFF
--- a/src/framework/button/AlertButton.jsx
+++ b/src/framework/button/AlertButton.jsx
@@ -1,0 +1,23 @@
+
+import React from 'react';
+import classNames from 'classnames';
+
+import Button from './Button.jsx';
+
+const AlertButton = (props) => {
+  const classes = classNames('button--alert', props.classes);
+
+  const extendedProps = Object.assign({}, props, {
+    classes,
+  });
+
+  return (
+    <Button {...extendedProps} />
+  );
+};
+
+AlertButton.TYPE = Object.assign({}, Button.TYPE);
+
+AlertButton.propTypes = Object.assign({}, Button.propTypes);
+
+export default AlertButton;
--- a/src/framework/button/AlertButton.jsx
+++ b/src/framework/button/AlertButton.jsx
@@ -1,6 +1,6 @@
 
-import React from 'react';
 import classNames from 'classnames';
+import React from 'react';
 
 import Button from './Button.jsx';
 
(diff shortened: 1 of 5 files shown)


src/framework/button/AlertButton.jsx
```diff
@@ -0,0 +1,23 @@
+
+import React from 'react';
+import classNames from 'classnames';
```
ahollesen: Order.

src/framework/button/_button.scss
```diff
@@ -71,6 +71,12 @@ $ctaButtonColor: #35d0a0;
  * 1. Fill in the background with the top gradient color to fix an IE11 bug
  *    where the transparent background color was showing through at the top.
  */
+.button--alert {
+  background: #f43434;
+  color: #ffffff;
+  border-color: #d03535;
+}
```
ahollesen: File could get cleaned up with e.g. order and not repeating `.button` but reusing parent.

src/guide/views/button/ButtonExample.jsx
```diff
@@ -79,6 +80,17 @@ export default class ButtonExample extends Component {
...
         </Example>
 
+        <Example title="AlertButton">
+          <Text>
+            This button lets the user perform a special action, like
+            &quot;Cancel&quot;. It&apos;s higher in the UI hierarchy than the
+            BasicButton.
+          </Text>
+          <AlertButton>
+            Cancel
```
ahollesen: Perhaps `Delete` makes more of an example for this button. And not sure of those descriptions are really needed but ok, just a suggestion.

src/guide/views/button/ButtonExample.jsx
```diff
@@ -79,6 +80,17 @@ export default class ButtonExample extends Component {
           </PrimaryButton>
         </Example>
 
+        <Example title="AlertButton">
+          <Text>
+            This button lets the user perform a special action, like
+            &quot;Cancel&quot;. It&apos;s higher in the UI hierarchy than the
```
muthan: I would say Cancel is not hard enough for an alert button. Maybe there are cases where you would use this button but in general Cancel would use a more neutral button because you wouldn't change anything,

I would suggest here to go with something like "Delete" where you have to be aware of the danger that is happening when you click that button.

PR #325, opened 2018-05-25
[SDX-1563] Create accordeon component:

mariacasancho:
- implementation done

### DIFF
--- a/src/framework/accordion/Accordion.jsx
+++ b/src/framework/accordion/Accordion.jsx
@@ -3,7 +3,8 @@ import React, {
   Component,
   PropTypes,
 } from 'react';
-import classNames from 'classnames';
+
+import AccordionItem from './../accordionItem/AccordionItem.jsx';
 
 export default class Accordion extends Component {
 
@@ -11,7 +12,7 @@ export default class Accordion extends Component {
     super(props);
 
     this.state = {
-      activeID: this.props.activeID ? this.props.activeID : 0,
+      activeID: !isNaN(this.props.activeID) ? this.props.activeID : null,
     };
 
     this.onTitleClick = this.onTitleClick.bind(this);
@@ -25,40 +26,33 @@ export default class Accordion extends Component {
     }
   }
 
-  classNameContent(index) {
-    return classNames('accordion-item__content', {
-      'accordion-item__content--active': index === this.state.activeID,
-    });
-  }
-
   render() {
     const wrapperStyle = {
-      maxHeight: this.props.maxHeight ? this.props.maxHeight : '250px',
-      width: this.props.width ? this.props.width : '210px',
+      width: this.props.width,
     };
 
     return (
-      <div className="accordion">
+      <div className="accordion" style={wrapperStyle}>
         {this.props.children.map((child, index) => (
-          <div key={index} style={wrapperStyle} className="accordion-item">
-            <div
-              className="accordion-item__title"
-              onClick={() => (this.onTitleClick(index))}
-            >
-              {child.title}
-            </div>
-
-            <div
-              className={this.classNameContent(index)}
-            >
-              {child.content}
-            </div>
-          </div>
+          <AccordionItem
+            index={index}
+            isActive={this.state.activeID === index}
+            key={index}
+            maxHeight={this.props.maxHeight}
+            onTitleClick={this.onTitleClick}
+            title={child.props.title}
+          >
+            {child.props.children}
+          </AccordionItem>
         ))}
       </div>
     );
   }
 }
+
+Accordion.defaultProps = {
+  width: '210px',
+};
 
 Accordion.propTypes = {
   activeID: PropTypes.number,
--- a/src/framework/accordion/Accordion.jsx
+++ b/src/framework/accordion/Accordion.jsx
@@ -12,16 +12,16 @@ export default class Accordion extends Component {
     super(props);
 
     this.state = {
-      activeID: !isNaN(this.props.activeID) ? this.props.activeID : null,
+      activeId: this.props.activeId,
     };
 
     this.onTitleClick = this.onTitleClick.bind(this);
   }
 
   onTitleClick(currentId) {
-    if (currentId !== this.state.activeID) {
+    if (currentId !== this.state.activeId) {
       this.setState({
-        activeID: currentId,
+        activeId: currentId,
       });
     }
   }
@@ -36,7 +36,7 @@ export default class Accordion extends Component {
         {this.props.children.map((child, index) => (
           <AccordionItem
             index={index}
-            isActive={this.state.activeID === index}
+            isActive={this.state.activeId === index}
             key={index}
             maxHeight={this.props.maxHeight}
             onTitleClick={this.onTitleClick}
@@ -55,7 +55,7 @@ Accordion.defaultProps = {
 };
 
 Accordion.propTypes = {
-  activeID: PropTypes.number,
+  activeId: PropTypes.number,
   children: PropTypes.array,
   maxHeight: PropTypes.string,
   width: PropTypes.string,
--- a/src/framework/accordion/Accordion.jsx
+++ b/src/framework/accordion/Accordion.jsx
@@ -1,0 +1,69 @@
+
+import React, {
+  Component,
+  PropTypes,
+} from 'react';
+import classNames from 'classnames';
+
+export default class Accordion extends Component {
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      activeID: this.props.activeID ? this.props.activeID : 0,
+    };
+
+    this.onTitleClick = this.onTitleClick.bind(this);
+  }
+
+  onTitleClick(currentId) {
+    if (currentId !== this.state.activeID) {
+      this.setState({
+        activeID: currentId,
+      });
+    }
+  }
+
+  classNameContent(index) {
+    return classNames('accordion-item__content', {
+      'accordion-item__content--active': index === this.state.activeID,
+    });
+  }
+
+  render() {
+    const wrapperStyle = {
+      maxHeight: this.props.maxHeight ? this.props.maxHeight : '250px',
+      width: this.props.width ? this.props.width : '210px',
+    };
+
+    return (
+      <div className="accordion">
+        {this.props.children.map((child, index) => (
+          <div key={index} style={wrapperStyle} className="accordion-item">
+            <div
+              className="accordion-item__title"
+              onClick={() => (this.onTitleClick(index))}
+            >
+              {child.title}
+            </div>
+
+            <div
+              className={this.classNameContent(index)}
+            >
+              {child.content}
+            </div>
+          </div>
+        ))}
+      </div>
+    );
+  }
+}
+
+Accordion.propTypes = {
+  activeID: PropTypes.number,
+  children: PropTypes.array,
+  maxHeight: PropTypes.string,
+  width: PropTypes.string,
+};
+
--- a/src/framework/accordion/Accordion.spec.jsx
+++ b/src/framework/accordion/Accordion.spec.jsx
@@ -1,0 +1,53 @@
+
+import { TestCaseFactory } from 'react-test-kit';
+import Accordion from './Accordion.jsx';
+
+describe('Accordion', () => {
+  describe('Props', () => {
+    const props = {
+      activeId: 0,
+      children: [{
+        title: 'test',
+        content: 'testing',
+      }],
+      maxHeight: '100px',
+      width: '100px',
+    };
+    describe('children', () => {
+      it('is rendered', () => {
+        const testCase = TestCaseFactory.create(Accordion, props);
+        expect(testCase.dom.textContent).toContain('test');
+      });
+    });
+
+    describe('activeId', () => {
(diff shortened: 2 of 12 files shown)


src/framework/accordion/Accordion.jsx
```diff
@@ -0,0 +1,69 @@
...
+  render() {
+    const wrapperStyle = {
+      maxHeight: this.props.maxHeight ? this.props.maxHeight : '250px',
+      width: this.props.width ? this.props.width : '210px',
+    };
+
+    return (
+      <div className="accordion">
+        {this.props.children.map((child, index) => (
+          <div key={index} style={wrapperStyle} className="accordion-item">
```
ahollesen: The `item` would already be the child of the `accordion` wouldn't it? Then `accordion__item`.

ahollesen: Order of attr.

src/framework/accordion/Accordion.jsx
```diff
@@ -0,0 +1,69 @@
...
+            <div
+              className="accordion-item__title"
+              onClick={() => (this.onTitleClick(index))}
+            >
+              {child.title}
+            </div>
+
+            <div
+              className={this.classNameContent(index)}
+            >
```
ahollesen: One liner.

src/framework/accordion/Accordion.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+    return (
+      <div className="accordion">
+        {this.props.children.map((child, index) => (
+          <div key={index} style={wrapperStyle} className="accordion-item">
+            <div
+              className="accordion-item__title"
+              onClick={() => (this.onTitleClick(index))}
+            >
+              {child.title}
```
ahollesen: `props.children` has to contain a specific structure. Should not be expected from client to implement correctly, rather create sub component for that specifies structure. Perhaps extract the whole accordion item.

src/framework/accordion/Accordion.spec.jsx
```diff
@@ -0,0 +1,53 @@
...
+        title: 'test',
+        content: 'testing',
+      }],
+      maxHeight: '100px',
+      width: '100px',
+    };
+    describe('children', () => {
+      it('is rendered', () => {
+        const testCase = TestCaseFactory.create(Accordion, props);
+        expect(testCase.dom.textContent).toContain('test');
```
ahollesen: Could check that both `title` and `content` is rendered.

src/framework/accordion/Accordion.spec.jsx
```diff
@@ -0,0 +1,53 @@
...
+      });
+    });
+
+    describe('activeId', () => {
```
ahollesen: Order of props tested.

src/framework/accordion/Accordion.spec.jsx
```diff
@@ -0,0 +1,53 @@
+
+import { TestCaseFactory } from 'react-test-kit';
+import Accordion from './Accordion.jsx';
+
+describe('Accordion', () => {
+  describe('Props', () => {
+    const props = {
+      activeId: 0,
```
ahollesen: Typo of prop. That the test works is a side effect of the constructor.

src/framework/accordion/Accordion.spec.jsx
```diff
@@ -0,0 +1,53 @@
...
+      children: [{
+        title: 'test',
+        content: 'testing',
+      }],
+      maxHeight: '100px',
+      width: '100px',
+    };
```
ahollesen: New line.

src/framework/accordion/Accordion.spec.jsx
```diff
@@ -0,0 +1,53 @@
+
+import { TestCaseFactory } from 'react-test-kit';
+import Accordion from './Accordion.jsx';
+
+describe('Accordion', () => {
+  describe('Props', () => {
```
ahollesen: Could start testing the structure of code as well (like we do in SEM). Let's talk about this point.

src/framework/accordion/Accordion.spec.jsx
```diff
@@ -0,0 +1,53 @@
...
+        )
+        .toEqual(1);
+      });
+    });
+
+    describe('maxHeight', () => {
+      it('is added', () => {
+        const testCase = TestCaseFactory.create(Accordion, props);
+        expect(
+          testCase.dom.getElementsByClassName('accordion-item')[0].style.cssText
```
ahollesen: Should use `testCase.first('.accordion-item')` instead.

src/framework/accordion/Accordion.jsx
```diff
@@ -0,0 +1,69 @@
...
+} from 'react';
+import classNames from 'classnames';
+
+export default class Accordion extends Component {
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      activeID: this.props.activeID ? this.props.activeID : 0,
```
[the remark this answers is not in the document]
ahollesen: And as for now there is no possibility to collapse the whole thing. Might be out of scope but would come into play when considering for the new targeting summary.

src/framework/accordion/Accordion.spec.jsx
```diff
@@ -0,0 +1,53 @@
...
+        const testCase = TestCaseFactory.create(Accordion, props);
+        expect(testCase.dom.textContent).toContain('test');
+      });
+    });
+
+    describe('activeId', () => {
+      it('only one is active', () => {
+        const testCase = TestCaseFactory.create(Accordion, props);
+        expect(
+          testCase.dom.querySelectorAll(
```
ahollesen: Could use `find` instead.

src/framework/accordion/Accordion.jsx
```diff
@@ -0,0 +1,69 @@
...
+  classNameContent(index) {
+    return classNames('accordion-item__content', {
+      'accordion-item__content--active': index === this.state.activeID,
+    });
+  }
+
+  render() {
+    const wrapperStyle = {
+      maxHeight: this.props.maxHeight ? this.props.maxHeight : '250px',
+      width: this.props.width ? this.props.width : '210px',
```
ahollesen: The `accordion-item` should have the `width` but rather the `accordion` itself. Or even the container around it so it could just take the full width.

src/framework/accordion/_accordion.scss
```diff
@@ -0,0 +1,44 @@
...
+    line-height: $lineHeight;
+    padding: 0 10px;
+  }
+
+  &__content {
+    background-color: #ffffff;
+    display: none;
+    flex: 1 1 auto;
+    line-height: $lineHeight;
+    opacity: 0;
```
ahollesen: Needed?

src/framework/accordion/_accordion.scss
```diff
@@ -0,0 +1,44 @@
...
+    overflow-y: auto;
+    padding: 5px 10px;
+
+    &--active {
+      animation: display-none-transition 1s both;
+      display: block;
+      opacity: 1;
+      -webkit-animation: display-none-transition 1s both; // scss-lint:disable VendorPrefix
```
ahollesen: Should not have vendor specifics in code base.

src/framework/accordion/_accordion.scss
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    @keyframes display-none-transition { // scss-lint:disable VendorPrefix
```
ahollesen: Don't think the disable is needed here.

src/framework/framework.js
```diff
@@ -1,5 +1,6 @@
 
 export * from './addOnControl/AddOnControl.jsx';
+export { default as Accordion } from './accordion/Accordion.jsx';
```
ahollesen: Order.

src/guide/views/accordion/AccordionExample.jsx
```diff
@@ -0,0 +1,48 @@
...
+    'sphere',
+    'vitality',
+  ];
+  const results = items.map((item, index) => (
+    <Text key={index}>{item}</Text>
+  ));
+  const arrayComponents = [];
+  arrayComponents.push(<Text>This is some text.</Text>);
+  arrayComponents.push(results);
+  return (<Page title={props.route.name}>
```
ahollesen: Generally some new line would help and here wrap `Page` to next line.

ahollesen:
Also background color (perhaps too dark, know it comes from the `Panel`) and borders (too thick) should be revisited.

muthan:
I also thought about the background color. But then matched it with the ones from SDX that I would use.
It is the same grey as used in the title rows of targeting, So I think its fine. Borders on the other hand I agree. Especially in between title rows it seems a bit to thick.

mariacasancho:
please review again

src/framework/accordion/Accordion.spec.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+        testCase.trigger('click', testCase.find(
+          '.accordion__item__title'
+        )[1]);
+
+        expect(testCase.find(
+          '.accordion__item__content'
+        )[1].getAttribute('class'))
+        .toContain('accordion__item__content--active');
+      });
```
ahollesen: Could check that click at `[0]` then one is active, click the `[1]` and then was active before and not anymore. And if that works as such merge both test cases for `activeId`.

src/framework/accordion/Accordion.spec.jsx
```diff
@@ -0,0 +1,80 @@
...
+      </AccordionItem>,
+      <AccordionItem key={1} title={'title2'}>
+        {'test2'}
+      </AccordionItem>,
+    ],
+    maxHeight: '100px',
+    width: '100px',
+  };
+
+  describe('DOM structure', () => {
```
ahollesen: Could also check `expect(testCase.dom.className).toContain('accordion');`

src/framework/accordionItem/AccordionItem.jsx
```diff
@@ -0,0 +1,56 @@
+
+import React, {
+  Component,
+  PropTypes,
+} from 'react';
+import classNames from 'classnames';
```
ahollesen: Order.

src/framework/accordionItem/AccordionItem.spec.jsx
```diff
@@ -0,0 +1,51 @@
...
+    describe('maxHeight', () => {
+      it('is added', () => {
+        const testCase = TestCaseFactory.create(AccordionItem, props);
+        expect(
+          testCase.dom.style.cssText
+        ).toContain('max-height: 100px');
+      });
+    });
+
+    describe('Title', () => {
```
ahollesen: `title`

src/framework/accordionItem/AccordionItem.spec.jsx
```diff
@@ -0,0 +1,51 @@
...
+        expect(testCase.dom.textContent).toContain('testing');
+      });
+    });
+
+    describe('maxHeight', () => {
+      it('is added', () => {
+        const testCase = TestCaseFactory.create(AccordionItem, props);
+        expect(
+          testCase.dom.style.cssText
+        ).toContain('max-height: 100px');
```
ahollesen: One liner.

src/framework/accordionItem/AccordionItem.spec.jsx
```diff
@@ -0,0 +1,51 @@
...
+          ).length
+        )
+        .toEqual(1);
+      });
+    });
+
+    describe('children', () => {
+      it('is rendered', () => {
+        const testCase = TestCaseFactory.create(AccordionItem, props);
+        expect(testCase.dom.textContent).toContain('testing');
```
ahollesen: Could then test on `props.children` instead.

src/framework/accordionItem/AccordionItem.jsx
```diff
@@ -0,0 +1,56 @@
...
+        <div className={this.classNameContent()}>
+          {this.props.children}
+        </div>
+      </div>
+    );
+  }
+}
+
+AccordionItem.defaultProps = {
+  height: '250px',
```
ahollesen: Should be `maxHeight `

src/framework/framework.js
```diff
@@ -1,4 +1,7 @@
 
+export { default as Accordion } from './accordion/Accordion.jsx';
+export { default as AccordionItem } from './accordionItem/AccordionItem.jsx';
```
ahollesen: Could possibly be exported through `Accordion`.

src/framework/accordion/Accordion.spec.jsx
mariacasancho: I tried to stick to one case per test. I think that only one item is active and that the clicked one is active are 2 different cases.

src/framework/accordion/Accordion.jsx
```diff
@@ -0,0 +1,63 @@
+
+import React, {
+  Component,
+  PropTypes,
+} from 'react';
+
+import AccordionItem from './../accordionItem/AccordionItem.jsx';
```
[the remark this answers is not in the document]
mariacasancho: as discussed, I prefer to keep it like it is

PR #327, opened 2018-06-18
SDX-1523 accordion component fixes

arijitdasgupta:
Made the accordion almost stateless, except AccordionItem knowing it's own state of open-ness. Refer to example source for more details.

### DIFF
--- a/src/framework/accordionItem/AccordionItem.jsx
+++ b/src/framework/accordionItem/AccordionItem.jsx
@@ -23,13 +23,13 @@ export default class AccordionItem extends Component {
     });
   }
 
-  onTitleClick() {
+  onTitleClick(e) {
     this.setState({
       isActive: !this.state.isActive,
     });
 
     if (this.props.onTitleClick) {
-      this.props.onTitleClick();
+      this.props.onTitleClick(e);
     }
   }
 
--- a/src/framework/accordionItem/AccordionItem.jsx
+++ b/src/framework/accordionItem/AccordionItem.jsx
@@ -10,12 +10,32 @@ export default class AccordionItem extends Component {
   constructor(props) {
     super(props);
 
-    this.onTitleClick = this.props.onTitleClick.bind(this);
+    this.state = {
+      isActive: props.isActive,
+    };
+
+    this.onTitleClick = this.onTitleClick.bind(this);
+  }
+
+  componentWillReceiveProps(nextProps) {
+    this.setState({
+      isActive: nextProps.isActive,
+    });
+  }
+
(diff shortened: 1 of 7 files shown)


src/framework/accordionItem/AccordionItem.jsx
```diff
@@ -10,12 +10,32 @@ export default class AccordionItem extends Component {
...
-    this.onTitleClick = this.props.onTitleClick.bind(this);
+    this.state = {
+      isActive: props.isActive,
+    };
+
+    this.onTitleClick = this.onTitleClick.bind(this);
+  }
+
+  componentWillReceiveProps(nextProps) {
+    this.setState({
```
ahollesen: Could do only when `this.state.isActive !== nextProps.isActive` to prevent too many re-renderings.

src/framework/accordionItem/AccordionItem.jsx
```diff
@@ -10,12 +10,32 @@ export default class AccordionItem extends Component {
...
+      isActive: nextProps.isActive,
+    });
+  }
+
+  onTitleClick(e) {
```
ahollesen: Perhaps call it `event`.

src/framework/accordionItem/AccordionItem.spec.jsx
```diff
@@ -14,14 +13,20 @@ describe('AccordionItem', () => {
     };
 
     describe('isActive', () => {
-      it('it has the proper class when it is active', () => {
+      it('is true then the component has proper class', () => {
```
ahollesen: I think we usually rather do `when true then sets something` or so...

src/framework/accordionItem/AccordionItem.spec.jsx
```diff
@@ -14,14 +13,20 @@ describe('AccordionItem', () => {
...
         const testCase = TestCaseFactory.create(AccordionItem, props);
-        expect(
-          testCase.find(
-            '.accordion__item__content--active'
-          ).length
-        )
-        .toEqual(1);
+        expect(testCase.find('.accordion__item__content--active').length)
+          .toEqual(1);
```
ahollesen: Could also do `expect(testCase.first('.accordion__item__content--active')).beBeDefined()` and `.not.toBe...` for the second case.

src/framework/accordionItem/AccordionItem.jsx
arijitdasgupta: Good idea!

arijitdasgupta:
PRA

PR #328, opened 2018-06-26
TF-392: Fix circular dependencies in Button and Kpi components

tjaniczek:
As `webpack` can't handle circular dependencies (see https://github.com/webpack/webpack/issues/1788) UI-Framework can't be imported to projects compiled with `webpack`/`babel`. 

This PR removes circular imports from two files: `Button.jsx` and `Kpi.jsx` and replaces them with separate export files (`index.jsx`) for all related components. This way child components (like `HollowButton`) can rely on parents code (e.g. `Button.TYPE`) as they are imported from `index.jsx` not `Button.jsx` file.

As far as I know all exports behave exactly the same way as previously, so this shouldn't introduce any regression. Also, all units test pass locally. Unfortunately as an external developer I can't test this with other apps using the framework, so before merging **please** test this thoroughly. 

Any questions please let me know!

### DIFF
--- a/src/framework/button/Button.jsx
+++ b/src/framework/button/Button.jsx
@@ -97,29 +97,3 @@ Button.propTypes = {
 };
 
 export default Button;
-
-// We need to export these components after the default export because they
-// depend on it.
-export {
-  default as AlertButton,
-} from './AlertButton.jsx';
-
-export {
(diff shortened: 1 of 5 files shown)


arijitdasgupta:
Looks good. Merging.

PR #330, opened 2018-07-13
Feature/sdx 1541 add edit functionality to card

mariacasancho:


src/framework/card/Card.jsx
```diff
@@ -1,52 +1,87 @@
...
+                  </Label>
+                </Tooltip>
+              </div>
+            </div>
+            <div className="card__footer--right">
+              <div className="card__footer--right__status">
+                <StatusDropdown
+                  onSelect={this.onSelectOption}
+                  options={this.dropdownItems}
+                  selectedOption={this.state.selectedOption}
```
arijitdasgupta: Maybe it's better to have the whole `StatusDropdown` as a prop passed into the component, and using `PropTypes.instanceOf` to only allow dropdown to be passed in as a prop. That way the card component don't need know anything about the callback, selectable values etc.

src/framework/statusDropdown/StatusDropdown.jsx
```diff
@@ -105,14 +121,17 @@ export default class StatusDropdown extends Component {
       name = this.optionToMetaDataMap[option].name.inactive;
     }
 
-    return [
+    const iconAvailable = !!this.optionToMetaDataMap[option].iconType;
```
arijitdasgupta: This check maps the `tick` to the icon that represents the action, and if there are no icon to it's left, no `tick` to it's right. 

Maybe it's slightly more semantic if we have a `showTick` property in the `optionToMetaDataMap`? What do you think? 

That way it's more explicit what the condition is all about. Right now it's basically a implicit relationship between having a icon to the right => having `tick` to the left.

src/framework/statusDropdown/StatusDropdown.jsx
```diff
@@ -105,14 +121,17 @@ export default class StatusDropdown extends Component {
...
+    return iconAvailable ? [
       <StatusDropdownOptionIcon
         key={0}
         type={type}
       />,
       <span key={1}>{name}</span>,
       selectedIcon,
-    ];
+    ] :
+    [<span key={1}>{name}</span>];
```
arijitdasgupta: Relating to my previous comment, this could be little more simplified.

src/framework/card/_card.scss
```diff
@@ -35,15 +35,36 @@
     }
 
     &__footer {
+      display: flex;
+      flex-direction: row;
+      height: 24px;
+      justify-content: center;
+      text-align: center;
```
ahollesen: I think those styles are a little overkill. `display: flex` could be fine and the text-alignment should only be applied to the dropdown content (or used somewhere else?). Also the specified heights here and in status doesn't seem to be needed.

src/framework/statusDropdown/StatusDropdown.jsx
mariacasancho: talking with Arijit we decided this is a bit out of scope of this ticket.

mariacasancho:
please review again

arijitdasgupta:
Maybe, as a future reference, maybe change the Card example to specifically mention that there are injected props as StatusDropdown.

src/framework/card/Card.jsx
```diff
@@ -1,55 +1,70 @@
...
+              <div className="card__footer--left__tooltip">
+                <Tooltip message={this.props.tooltipText} width="150px">
+                  <Label className="card__footer__label card__text--green">
+                    {this.props.hightlightText}
+                  </Label>
+                </Tooltip>
+              </div>
+            </div>
+            <div className="card__footer--right">
+              <div className="card__footer--right__status">
```
ahollesen: This div should probably be part of passed prop as well.

src/framework/card/Card.jsx
```diff
@@ -1,55 +1,70 @@
...
-  );
-};
+      </Box>
+    );
+  }
+}
 
 Card.propTypes = {
   description: PropTypes.string,
+  footerRight: PropTypes.element,
```
ahollesen: Tests missing that prop is rendered.

src/framework/card/_card.scss
```diff
@@ -35,15 +35,33 @@
...
+          display: flex;
+          height: 20px;
+        }
+      }
+
+      &--right {
+        flex-basis: 50%;
+        line-height: 1.5;
+
+        &__status .statusDropdown {
```
ahollesen: Should come from client CSS.

src/framework/card/_card.scss
```diff
@@ -60,4 +78,9 @@
       color: $ultraLightGrey;
     }
   }
+
+  .statusDropdownOptionList {
+    bottom: -8px;
+    top: inherit;
+  }
```
ahollesen: This as well.

src/framework/statusDropdown/StatusDropdown.jsx
```diff
@@ -105,14 +121,19 @@ export default class StatusDropdown extends Component {
...
 
-    return [
-      <StatusDropdownOptionIcon
-        key={0}
-        type={type}
-      />,
-      <span key={1}>{name}</span>,
-      selectedIcon,
-    ];
+    const iconAvailable = !!this.optionToMetaDataMap[option].iconType;
```
ahollesen: Technically no need to cast to boolean.

PR #332, opened 2018-08-06
Card component centered picture fix

arijitdasgupta:


alanbasse (approved):
Don't forget also to apply within pre package deal read modal.

PR #340, opened 2018-10-02
Adjusted dependencies to enable installation with Node 8 and npm 5

jenslonkowski:


package.json
```diff
@@ -74,6 +72,7 @@
     "react-transform-hmr": "1.0.1",
     "reconnectingwebsocket": "1.0.0",
     "run-sequence": "1.1.2",
+    "selenium-standalone": "6.15.3",
```
jaideepsingh: Why have we added this?

jenslonkowski: Because the "gulp-tasks" package requires this dependency directly in "node_modules". Previously it was enough to have the dependency in "gulp-tasks" but newer versions of npm create a hierarchy inside "node_modules" so this didn't work anymore.

jenslonkowski:
Good point @ahollesen - I added it.

PR #345, opened 2018-11-20
Update node version

arijitdasgupta:
This pull request makes sure the build work with node v11.x.x


arijitdasgupta:
Please review again.

arijitdasgupta:
@leofaster IMHO we should not specify that in our README, since it's an `npm` feature, and the user of a `>5.8` NPM can be expected to be aware of that.

PR #354, opened 2019-08-22
RadioButtons

skandarhayek:


src/framework/radioButtons/RadioButtons.jsx
```diff
@@ -0,0 +1,81 @@
...
+  className: PropTypes.string,
+  elements: PropTypes.arrayOf(PropTypes.shape({
+    label: PropTypes.string,
+    value: PropTypes.any,
+  })).isRequired,
+  onSelect: PropTypes.func.isRequired,
+  selectedElement: PropTypes.shape({
+    label: PropTypes.string,
+    value: PropTypes.any,
+  }),
```
mariacasancho: you need to add name here. I get an error: 
47:30  error  'name' is missing in props validation

src/framework/organizationSwitcher/_organizationSwitcher.scss
```diff
@@ -9,7 +9,7 @@ body {
 }
 
 .organizationSwitcher {
-  border: 1px solid #d4dadc;
+  border: 1px solid $borderGrey;
```
mariacasancho, line 12: order

src/framework/radioButtons/RadioButtons.spec.jsx
```diff
@@ -0,0 +1,53 @@
...
+        const testCase = TestCaseFactory.create(RadioButtons, props);
+        expect(
+          testCase.find('.radioButtons--element').length
+        ).toEqual(2);
+      });
+    });
+  });
+
+  describe('Actions', () => {
+    describe('upon clicking of a option', () => {
```
mariacasancho: should we have a test to check that contains the active class?

mariacasancho (changes requested):
just small comments, good job!

src/framework/radioButtons/RadioButtons.jsx
arijitdasgupta: Thanks! Good spot. Done.

src/framework/radioButtons/RadioButtons.jsx
```diff
@@ -0,0 +1,101 @@
...
+    this.setState({
+      selectedOption: element,
+    });
+    this.props.onSelect(element.value);
+  }
+
+  render() {
+    const onSelectCreator = value => () => this.onSelectOption(value);
+
+    const radioElements = this.props.elements.map((element, index) => {
```
DarioG: What about creating another RadioElement component?

src/framework/radioButtons/RadioButtons.jsx
```diff
@@ -0,0 +1,101 @@
...
+
+  onSelectOption(element) {
+    this.setState({
+      selectedOption: element,
+    });
+    this.props.onSelect(element.value);
+  }
+
+  render() {
+    const onSelectCreator = value => () => this.onSelectOption(value);
```
DarioG: If we create the RadioElement component we wouldnt have to create this hacky wrapper

src/framework/radioButtons/RadioButtons.jsx
```diff
@@ -0,0 +1,101 @@
...
+      // https://reactjs.org/docs/react-component.html#componentdidmount
+      // eslint-disable-next-line react/no-did-mount-set-state
+      this.setState({
+        selectedOption: this.props.selectedElement
+      });
+    }
+  }
+
+  componentDidUpdate() {
+    if (this.state.selectedOption !== this.props.selectedElement) {
```
DarioG: You have 2 methods duplicated here.... this is a common symptom of not having a single source of truth....in this case for the selectedElement....If you ask me i would remove completely the state of the component and leave it up to the client to keep track of the selected element......

DarioG (commented):
I added some comments....besides I think the test suite is missing a few cases that, in my opinion, should be tested

src/framework/radioButtons/RadioButtons.jsx
arijitdasgupta: This was done this way to keep consistency with other components in the UI Framework, such as `Dropdown`. But totally I agree with you in general. 👍

mariacasancho (approved):

PR #398, opened 2022-08-02
Sem 595 update outdated components

alexpagri:
Added the "data-testid" field in components.
Replaced outdated "unsafe" React components with newer alternatives.


alexpagri:
So, I used Node 11 and some legacy peer deps argument for npm install to get it to a partially working state. This was enough to run all tests and check their output. I will push some changes. The tests ran successfully but there are some warnings that appeared because of the changes I did.

